PR #8, opened 2017-06-08
Enables prefetching images

fjcaetano:
I have created a method that receives a string or an array of strings which prefetches and caches these urls for later use.


DylanVann:
Thanks for the PR. I modified the API a bit:

- renamed to preload
- takes sources in the same format as the source prop
- only takes an array of sources

PR #34, opened 2017-08-04
Remove override of createJSModules

exPHAT:
Since of recently in master the unused createJSModules has been removed and ReactPackage's no longer contain the definition of createJSModules. There for it must be removed in order to not yield a compile error. This is breaking change.

facebook/react-native@ce6fb33


dwillmc:
Please merge
Confirmed fixes annoying breaking-change in RN for all existing Android native modules

DylanVann:
Awesome, thank you!

PR #36, opened 2017-08-06
Remove overriden createJSModules

mortezaalizadeh:
Since of recently in master the unused createJSModules has been removed and ReactPackage's no longer contain the definition of createJSModules. There for it must be removed in order to not yield a compile error. This is breaking change.

### DIFF
--- a/android/src/main/java/com/dylanvann/fastimage/FastImageViewPackage.java
+++ b/android/src/main/java/com/dylanvann/fastimage/FastImageViewPackage.java
@@ -18,6 +18,11 @@ public class FastImageViewPackage implements ReactPackage {
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


exPHAT:
Duplicate of https://github.com/DylanVann/react-native-fast-image/pull/34 ?

mortezaalizadeh:
No, it is not. I only removed the @override to keep it backward compatible with those still using older version of react-native. It is up to you guys how you want to move forward. Cheers

DylanVann:
Thank you @mortezaalizadeh . Considering the pace of React Native development I'm not going to be trying to support older versions. People using older versions of RN can use older versions of this module. Thanks for the contribution!

PR #38, opened 2017-08-08
Fix backward compatibility for RN < 0.47

Amurmurmur:


DylanVann:
I'm not really interested in keeping this backward compatible unless there's a good argument for it. This isn't the first breaking change in React Native and probably won't be the last. It's best to just keep up to date with it.

Amurmurmur:
@DylanVann Fair enough, but just to make it clear, there are bunch of react-native modules that haven't been updated yet to support an Android breaking change for RN 0.47, including react-native-maps, react-native-config etc.. all of them are kind often used and well known, and it takes time to update those. 
While you can benefit and use new features from the updates you make to react-native-fast-image so that once you upgrade to the latest RN it happens in a smooth way 👍

Clowning:
Thank @Amurmurmur I needed that! 🎉

JPig:
i'm all about encouraging using the latest release, but the fact of the matter is not all packages/modules are as quick as each other to update/upgrade. So if some where to upgrade to 0.47 they could have 10 other packages lagging behind and have to fork PR's just for their dependencies. IMO this PR is a good idea, especially due to the simplicity of it.

DanGDroid:
dew to this issue i upgraded RN to from 0.45 to 0.47.
but still cannot build Android

ellman121:
> It's best to just keep up to date with it
> @DylanVann 

I feel like this is really flawed logic, especially for an app developer.

You wouldn't make an Android app and say "We only support Android 7+ because we want to encourage people upgrade their devices."  You're leaving out like 85% of the market

I would be willing to bet that if you did a survey of all of the RN apps out there, most of them would not be RN 0.47+.  It is silly to leave out a single method in a single java class because you want to force people who use your library to upgrade, when they may have valid reasons for not doing so.

matt-oakes:
I agree that this should be merged in as it's only a single method difference. The main reason we have for staying on 0.46 is that in 0.47 and above systrace is broken, which makes debugging performance issues almost impossible.

jmcneese:
i agree that this should be merged too, as it has no detrimental impact afaik.

hkung77:
I also agree, this is poor practice and it should be backwards compatible
I'll give an good real life example for all those web devs. 
"Unfortunately we still gotta support IE sometimes.."

DylanVann:
@Amurmurmur Thank you for the contribution, sorry I'm not merging it. I have added a note that `react-native-fast-image` will only aim to support the latest release of `react-native`

You're not required to update `react-native-fast-image`, just as you are not required to update `react-native`.

Only supporting the latest version of `react-native` makes this project easier to maintain and test.

@matt-oakes It does suck the systrace is broken on recent versions of RN, it seems like the core team is taking a look at it, hopefully it will be fixed soon.

Sorry everyone, not merging. Less code, less problems.

matt-oakes:
Thanks for taking the time to lay out the reason for the decision! 👍

Issue #37, opened 2017-08-08
Getting a wired behavior  bug with activity indicator  


AlmogRnD:
When I load my app and if go to my photo tabs which loads images from my redux store activity indicator is showing but does not go away attached gif showing the issue 
![fast-image-issue](https://user-images.githubusercontent.com/169507/29064072-b4a5d67e-7bdc-11e7-8abb-2968316c09c3.gif)

I don't think it's getting the correct events

DylanVann:
Make sure you rebuild the app if you just updated. I'll try out something like this and see if I have any issues.

AlmogRnD:
Yes I rebuild the app but still getting this issue

kesha-antonov:
Same here

ragnorc:
+1

erickjth:
+1

DannyvanderJagt:
Same here. I tried to debug it some more. 

These are the events that are dispatched when: 

**I load the image for the first time:**
- `onLoadStart` - once
- `onProgress` - multiple times
- `onLoad`
- `onLoadEnd`
- **the indicator disappears**

**When I reload the app and view the image:** - ***This image is loaded from cache I assume***
- `onLoadStart` - once
- **the indicator stays on screen**

**Strange behavior:**
- When I add the `onLoadEnd` event with an empty handler like so: `onLoadEnd={() => {}}` the indicator does disappear when I re-view the image.

**Proposed solution**
Maybe it is a good idea to make the api/event dispatching consistent by also firing the `onLoadEnd`, and the `onLoad` when appropriate, event when the image is loaded from cache. These changes will also benefit components and issues like: https://github.com/oblador/react-native-image-progress/issues/55


Let me know what your opinions are so we can fix this :)

ragnorc:
Any update on this?  @DylanVann

srameshr:
@DylanVann Yesh this still exists. Especially on IOS 9, the spinner wont just go away even if I add the empty onLoadEnd{() => {}}.
P

pavanmehta91:
@AlmogRnD  Try this. Somebody make a PR
https://github.com/DylanVann/react-native-fast-image/pull/74/files


Also can you suggest something on how to implement placeholder images (on similar lines as loading spinner)

simpleblack:
my solution

```javascript
<FastImage
  key={source && source.uri}
  source={source}
  style={StyleSheet.absoluteFill}
  onLoadStart={() => this.onLoadStart()}
[...]

PR #74, opened 2017-10-19
iOS: Ensure onLoad and onLoadEnd get called

danhodos:
Setting properties from JS to native components does not guarantee any
order (as far as I know). Because of this, `setSource` can get called
and complete before `setOnFastImageLoad` and `setOnFastImageLoadEnd`,
which results in the callbacks never firing.

This change implements a simple check for these callback properties
getting set after the image has loaded, and fires them at that time.


danhodos:
This might possibly fix #37.

dotansimha:
@DylanVann @danhodos 
Confirm that it solved issue with `react-native-image-progress`.

ovaris:
Please merge!

ovaris:
@DylanVann Any news on this, we really need onLoadEnd to be called also for cached imgs?

ovaris:
@danhodos I'm not able to build this on xcode (9.1):
```
node_modules/react-native-fast-image/ios/FastImage/FFFastImageView.h:3:9: fatal error: 'SDWebImage/UIImageView+WebCache.h' file not found
#import <SDWebImage/UIImageView+WebCache.h>
```

shinyford:
Hi @DylanVann - I'm having problems with callbacks not firing on iOS (works fine on Android) and it looks like this may fix it. Any chance that it could be merged in? Or are there other issues?

brunolemos:
@shinyford while it is not merged, you can apply it using [patch-package](https://github.com/ds300/patch-package)

patrickkempff:
@DylanVann it would be really nice to have this merged

varungupta85:
Verified that the fix worked for me.

DylanVann:
@danhodos Thank you for fixing this 😃 

Works great. I added an example of usage with `react-native-image-progress` as well.

progress

All these properties are really tied together, I've considered just setting everything in one object:

```jsx
<Image source={{ uri: 'www.dfsfd.com/thing.png', onLoadEnd: () => console.log('loaded') }} />
```

That could potentially solve all these issues. It's sort of a weird API though, and it would break compatibility with `react-native`'s `Image`.

DylanVann:
Sorry I messed up a rebase here. The actual commit is: https://github.com/DylanVann/react-native-fast-image/commit/d9f729915486665d9aad1f1febff5348ab3ab069

PR #75, opened 2017-10-20
use application context to avoid deallocated view error

wz366:
When split screen on Android, view can be reconstructed; therefore, using view context here can cause crashes on Android. Using application context avoids accessing context of a deallocated view.


wz366:
@wkoutre It actually works without force command :)

wkoutre:
@wz366 Not for me it doesn't ;)

wz366:
@wkoutre what is your OS?

wkoutre:
macOS 10.13

DylanVann:
Awesome! Thank you!

PR #80, opened 2017-11-08
local image not showing up on Android device.

tianling536:
fixed local image not showing up on Android device.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,8 @@
     "ios",
     "android",
     "utils",
-    "FastImage.js"
+    "FastImage.js",
+    "react-native-fast-image.podspec"
   ],
   "main": "FastImage.js",
   "directories": {


1c7:
why this still isn't merge into master branch?

shekharskamble:
any updates on this? I really need to display local images using FastImage

DylanVann:
Haven't had time to review this.

It's very trivial to write your own wrapper to do this. Also to clarify `react-native-fast-image` does not display local images, this PR is just to enable a fallback to `Image`. If we did want to really support showing local images I think we'd have an issue in that `SDWebImage` doesn't support that (see: https://github.com/rs/SDWebImage/issues/697#issuecomment-289838087).

DylanVann:
It might be possible actually, I will review this and release a version with the fallback. Then look into it as a new feature. I just want to clarify that this library as it is, or with this PR, will not improve performance when showing local images.

DylanVann:
I merged in your other PR that only made the change for supporting local images. CocoaPods support is now also included.

Thanks for the contribution!

PR #82, opened 2017-11-14
Upgrade android to Glide v4 

chitezh:


superburrito:
@chitezh Are you planning to merge this soon?

pine:
Glide v3 is too old. I’m in trouble 😔

chitezh:
@DylanVann Could you review this?

cyclops24:
@DylanVann Any news about this PR?

TilWs:
@DylanVann will this be implemented?

DylanVann:
@chitezh Thanks for the PR!

I tried to merge it in but had issues with the new Glide depending on newer support libraries. I had to revert it in the end.

You may know better than I do how to resolve the issues.

It needs to be able to be installed in the example application which is produced from `react-native init`.

DylanVann:
`feat/glide-4` is mostly working now. The only issue is I had to use `MyAppGlideModule` which uses the "generated API". If I don't use this generated API than `registerComponents` is never called, so the progress reporting doesn't work.

According to the glide documentation you should not include an `AppGlideModule` in a library because an app should only have one.

I'm not sure what the resolution to this is, because we need to be able to provide a module and to use the provided module.

PR #103, opened 2017-12-12
Added resource width and height information to the onFastImageLoad an…

jeremyclee:
…d onFastImageLoadEnd events

Instead of an empty map, it may be useful to have some information about the newly loaded resource. 
Since a user doesn't know the dimensions of a remote resource at load time, providing the dimensions on load end allows the user to adjust the view accordingly if required.


jeremyclee:
making a new PR from a specific branch

PR #105, opened 2017-12-13
Fix documentation for default resize mode

pehagg:
Documentation wrongly states that 'contain' is the default resize mode, when it in fact it is 'cover'.


DylanVann:
Awesome, thanks 👍 .

PR #140, opened 2018-02-02
Support to onReadDimensions

ararog:
Support to onReadDimensions to allow retrieve the original image dimensions and perform better layouting on the app which isn't aware of image height and width.


DylanVann:
@ararog Thank you for working on this. I've merged the other implementation. I think it makes sense to send the dimensions in `onLoadEnd`.

PR #169, opened 2018-03-19
fixed cannot display local image on android

tianling536:


DylanVann:
I'm looking into supporting local images on android and iOS, but this is a good fix until that's implemented. Thank you 👍

PR #172, opened 2018-03-27
Added null check on  getCurrentActivity()'s return value

Kouznetsov:
We noticed a lot of crashes because of getCurrentActivity() returning null. This PR contains a simple null check to avoid crashing.


LuigiQuarta:
I've encountered the same problem on my project and this fix resolve it. Thank you.

ifsnow:
@DylanVann Are you planning to merge this PR. I'm sure this crash is occurring. I hope this problem is resolved.

Kouznetsov:
Meanwhile, I made a fork with the fix applied. https://github.com/Kouznetsov/kh-rn-fast-image

DylanVann:
Looks reasonable. I guess in this case the preloading just won't work though. That's unfortunate, better not to crash though.

DylanVann:
Thanks @Kouznetsov

PR #179, opened 2018-04-06
Adding instructions to manually link.

Meandmybadself:
Adding instructions to manually link the module in the event that automatic linking isn't possible.


DylanVann:
Yeah we should probably have these, too bad `react-native link` doesn't always work. Thank you!

Issue #181, opened 2018-04-09
Dimensions not being passed to onLoad()


macavity23:
My onLoad handler is not receiving dimensions for the loaded image. Using code from the example:
```
<FastImage style={styles.photo}
           source={{uri: url}}
           onLoad={e => console.log(e.nativeEvent.width, e.nativeEvent.height)} />
```
This logs 'undefined, undefined' to the console. e.nativeEvent appears to have only one key, `target` (int).

The image is displaying fine. This is using 4.0.0 on the iOS simulator.

pingxiaodou:
The same issue

iorikingdom:
+1 
it seem working when using android ,but problem with ios

iorikingdom:
I found the point 
Because the _onFastImageLoad called without any params
``
                                NSDictionary* params = @{
                                                         @"width":[NSNumber numberWithDouble:image.size.width],
                                                         @"height":[NSNumber numberWithDouble:image.size.height]
                                                         };
                                fastImageLoadParams = params;

and 

        _onFastImageLoad(fastImageLoadParams);

will fixed it...

ligen52:
same issue+1 with iOS;
when it's first time ,the width and height is returned.
but when it loaded from cache ,no correct result returned.

DylanVann:
Yeah because the properties from JS can be set in a different order the loading of the image may finish before the callbacks are set.

Out of curiousity, what do you use the dimensions of the image for anyways?

Closing the issue, thanks for the PR @ligen52 .

macavity23:
To satisfy your curiosity, I need the image dimensions to calculate animation interpolations. Now I can use FastImage and my images won't flicker on load any more!

Thanks @ligen52 for the fix & @DylanVann for merge 🙏

manithin:
> Yeah because the properties from JS can be set in a different order the loading of the image may finish before the callbacks are set.
> 
> Out of curiousity, what do you use the dimensions of the image for anyways?
> 
> Closing the issue, thanks for the PR @ligen52 .

Is there any ways to solve this? In my case, I have a slideshow with swipable images. When the first image is loaded, onload returns me the correct dimensions. But from next image onwards(when I swipe on the component, it loads the next image), it just returns the dimension of the first loaded image.
I use the dimensions from the call back to adjust the size of the slideshow but now that does not work.

manithin:
> > Yeah because the properties from JS can be set in a different order the loading of the image may finish before the callbacks are set.
[...]

PR #182, opened 2018-04-16
fix bug :https://github.com/DylanVann/react-native-fast-image/issues/181

ligen52:
Dimensions not being passed to onLoad() #181


ligen52:
The ` _onFastImageLoad` is a little late to set when it loads the image from cache. It would cause the  size param not passed.

We need to record the size params and the put the param into the callback.

DylanVann:
Nice, this is the right idea.

PR #192, opened 2018-04-24
Using SDK Version variables from root project

rayronvictor:
Instead of assuming the `compileSdkVersion`, `targetSdkVersion`, etc, read it from the root project.
Default `compileSdkVersion` and `targetSdkVersion` to the latest versions.

Android Target API Level 26 will be required in August 2018.
https://android-developers.googleblog.com/2017/12/improving-app-security-and-performance.html
And the React Native team is already working on this:
facebook/react-native#17741
facebook/react-native#18095

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -9,18 +9,26 @@ buildscript {
 
 apply plugin: 'com.android.library'
 
+def _ext = rootProject.ext
+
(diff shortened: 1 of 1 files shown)


DylanVann:
Looks great, thank you!

PR #201, opened 2018-05-04
Upgrade to glide 4.

DylanVann:
This PR will upgrade the glide version for this package to Glide 4.


DylanVann:
Update on this: https://github.com/bumptech/glide/issues/2962#issuecomment-387953418

So I guess maybe users will have to be instructed to add `MyAppGlideModule` to their java code. That's unfortunate. We're also going to have to ask them to add the google maven repository already. 😢

DylanVann:
I could publish two versions of this package... one with an `AppGlideModule` that would work for most people, and one without for people using their own `AppGlideModule.

DylanVann:
nvm, there's a better way, conditionally excluding the class. documentation on this has been added.

PR #205, opened 2018-05-11
Show local images.

DylanVann:
Lots of people want it. It would also be a good step towards being independent from `Image` instead of using it as a fallback. Should make things more consistent.

This PR will be merged when both Android and iOS have support for all common local image usages:

- `https://`
  - [x] iOS
  - [x] Android
- `data:image/png;base64`
  - [x] iOS
  - [x] Android
- `file:///`
  - [x] Android
  - [x] iOS
- `photos://`
  - [ ] iOS
  - [ ] Android
- `content://`
  - [x] Android
- `assets-library://`
  - [ ] iOS
  - [ ] Android
- photos framework images
  - [ ] iOS
- whatever including a native image looks like
  - [ ] iOS
  - [ ] Android


n1ru4l:
Does this also support base64 encoded images?

DylanVann:
@n1ru4l I don't think it does. I could really use a list of all the types of image people might want to use to test this.

So far I can think of:
- `https://`
- `data:image/png;base64`
- `file://`
- `photos://`
- `assets-library://`
- photos framework images
- whatever including a native image looks like

So far I can only say for sure this works for `file://` packaged by React Native. It won't be merged until all types are supported and android is also working.

n1ru4l:
@DylanVann Okay I was just curios because I am currently using a base64 encoded low-res image for progressive image loading.

I can help you testing!

DylanVann:
Close enough, we'll do it live!

DylanVann:
@n1ru4l If you want to test this out it's on master now. It does work for `base64` images now.

I'm not sure it'll get much better performance in that use case (it may even get worse performance). The primary benefit of this module is better caching of remote images.

Definitely make sure your base64 image is small because the decoding is synchronous.

PR #213, opened 2018-05-28
useLocalImage: Expose & rename the useful function

belens:
I am using your handy function check to see when the image source is external and if it is, fade it in, would be cool if I can just import it.


DylanVann:
Hey @belens sorry I couldn't get to this sooner.

This function is probably going to be removed imminently and local images will be handled by `react-native-fast-image`.

If you want to use it feel free to copy it, make it into an npm module or anything you want.

PR #98, opened 2017-12-04
Fix memory leaks and add support border radius on android

chaitanya0bhagvan:
The Android implementation makes use of static variables that were causing memory leaks. Code refactor and use of WeakHashMap has the fixed memory leak due to this library.

Add support for border-radius on Android.

PR #214, opened 2018-06-01
[Android] Fixed memory leak

patrickkempff:
I was playing with [leakcanary](https://github.com/square/leakcanary) and noticed some memory leaks on Android. This PR fixes some of them. Credits go to @chaitanya0bhagvan for fixing those already in PR #98. I mainly did split this fix from the border-radius implementation in the same PR. 

At the moment the Android implementation uses static constants which never get releaed. Also got rid of HashMap in favor of WeakHashMap.

### DIFF
--- a/android/src/main/java/com/dylanvann/fastimage/FastImageOkHttpProgressGlideModule.java
+++ b/android/src/main/java/com/dylanvann/fastimage/FastImageOkHttpProgressGlideModule.java
@@ -17,6 +17,7 @@ import com.facebook.react.modules.network.OkHttpClientProvider;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.HashMap;
+import java.util.WeakHashMap;
 import java.util.Map;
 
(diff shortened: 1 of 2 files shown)


patrickkempff:
Halted until https://github.com/DylanVann/react-native-fast-image/pull/214 is merged

DylanVann:
Awesome! Thanks so much ✨ 

Currently `master` of this repo is the code for what will be the next major release. It will update the `Glide` version and enable loading local images with `FastImage`. Just checking you found these bugs on `master` not on the published version? (I wouldn't be surprised, just double checking)

patrickkempff:
@DylanVann awesome!

I noticed this leak first on `4.0.14` but I am confident it is also present on `master`. Unfortunately I am not in the position to confirm (test) at the moment.

On a side note, i have patched 4.0.14 with this fix to production (an app with over 2 million downloads and thousands of new users per day) without problems.

PR #215, opened 2018-06-04
Use <ImageBackground> if it uses local image and has children

morishin:
## Problem
`<FastImage>` cannot show its child components when it shows a local image because it uses just `<Image>` instead.

## Solution
I made `<FastImage>` use `<ImageBackground>` instead of `<Image>` if it has its child components.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -8,6 +8,7 @@ import {
     requireNativeComponent,
     ViewPropTypes,
     StyleSheet,
+    ImageBackground,
 } from 'react-native'
 
 const resolveAssetSource = require('react-native/Libraries/Image/resolveAssetSource')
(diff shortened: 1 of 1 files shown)


bfdevelop:
@morishin  Didn't it work as ImageBackground before?
I am using it as image background it works ok with all images, but it does not work only with images from Amazon s3.

DylanVann:
@morishin Thanks for the contribution. It will probably not be merged though because in the next major version `react-native-fast-image` will load local images without falling back to `Image`, so that code path won't be used.

WIP is on this PR: https://github.com/DylanVann/react-native-fast-image/pull/205

morishin:
😉👍

PR #216, opened 2018-06-04
Specify type definition file path

morishin:
We should specify the type definition file (`index.d.ts`) path in package.json.

TypeScript document says as follows.

>Also note that if your main declaration file is named index.d.ts and lives at the root of the package (next to index.js) you do not need to mark the "types" property, though it is advisable to do so.

https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html


DylanVann:
Awesome, thanks!

PR #231, opened 2018-06-26
Changes to use supportLibVersion

younes200:
Changes to use supportLibVersion instead of compileSdkVersion to  keep consistent with other modules (react-native-maps etc..) and avoid behavior conflicts when compileSdkVersion 26+

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -13,6 +13,7 @@ def _ext = rootProject.ext
 
 def _reactNativeVersion = _ext.has('reactNative') ? _ext.reactNative : '+'
 def _compileSdkVersion = _ext.has('compileSdkVersion') ? _ext.compileSdkVersion : 23
+def _supportLibVersion = _ext.has('supportLibVersion') ? _ext.supportLibVersion : 23
 def _buildToolsVersion = _ext.has('buildToolsVersion') ? _ext.buildToolsVersion : '23.0.1'
 def _minSdkVersion = _ext.has('minSdkVersion') ? _ext.minSdkVersion : 16
 def _targetSdkVersion = _ext.has('targetSdkVersion') ? _ext.targetSdkVersion : 22
@@ -55,7 +56,7 @@ repositories {
 dependencies {
     //noinspection GradleDynamicVersion
     compile "com.facebook.react:react-native:${_reactNativeVersion}"
-    compile "com.android.support:support-v4:${_compileSdkVersion}"
+    compile "com.android.support:support-v4:${_supportLibVersion}"
     compile("com.github.bumptech.glide:glide:${_glideVersion}") {
         exclude group: "com.android.support"
     }


compojoom:
@DylanVann - can we merge this? I had to do the same in order to get it working in my project.

n1ru4l:
@younes200 @compojoom I am getting the following error when trying to compile the `apk`

```
Android SDK Build Tools 27.0.3 will be used.
To suppress this warning, remove "buildToolsVersion '25.0.2'" from your build.gradle file, as each version of the Android Gradle Plugin now has a default version of the build tools.

/Users/groovecat/.gradle/caches/transforms-1/files-1.1/appcompat-v7-27.1.0.aar/ecef75a7c9bcbb18343a162dc71ec277/res/values-v24/values-v24.xml:3:5-157: AAPT: Error retrieving parent for item: No reso
urce found that matches the given name 'android:TextAppearance.Material.Widget.Button.Borderless.Colored'.

/Users/groovecat/.gradle/caches/transforms-1/files-1.1/appcompat-v7-27.1.0.aar/ecef75a7c9bcbb18343a162dc71ec277/res/values-v24/values-v24.xml:4:5-135: AAPT: Error retrieving parent for item: No resource found that matches the given name 'android:TextAppearance.Material.Widget.Button.Colored'.

/Users/groovecat/.gradle/caches/transforms-1/files-1.1/appcompat-v7-27.1.0.aar/ecef75a7c9bcbb18343a162dc71ec277/res/values-v26/values-v26.xml:13:5-16:13: AAPT: No resource found that matches the given name: attr 'android:keyboardNavigationCluster'.


FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':react-native-fast-image:verifyReleaseResources'.
> com.android.ide.common.process.ProcessException: Failed to execute aapt
```

Could this be related to this pull request?

younes200:
@n1ru4l yes, it's related. You can use my fork to confirme it : https://github.com/InteractiveObject/react-native-fast-image

raulmt:
This is the PR that should be merged imo, since #247 doesn't really solve the problem. RN 0.56 started using these `ext` variables in the project level `build.gradle`, and they added `supportLibVersion` too, which is used in the app's `build.gradle` to decide which version of appcompat-v7 to use. So I think it makes sense to use that same variable for this other support lib this package uses, right?

In my case, I have `compileSdkVersion` 27, so the compilation was failing because `27.+` gets resolved to 27.1.1, and that support-v4 version seems to be incompatible with latest npm version of this package currently (4.1.14). But since I have `supportLibVersion` set to 27.0.2, and that version seems to be compatible, this PR also fixes this problem I was having.

DylanVann:
Thanks for the contribution!

Looks like this change has already been made on master so I'm going to close this PR. Thank you though.

PR #252, opened 2018-07-24
little improvement over too many "def"

yeomann:
just added a function which acts like a ternary with fallback option. Hence, less cluttered "def" variables  
additionally, changed the SDK values from 23 to 26 as per new changes from react-native and Android

Android Target API Level 26 will be required in August 2018.
https://android-developers.googleblog.com/2017/12/improving-app-security-and-performance.html

And the React Native team is already working on this:
facebook/react-native#18095
facebook/react-native#17741

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -9,24 +9,19 @@ buildscript {
 
 apply plugin: 'com.android.library'
 
-def _ext = rootProject.ext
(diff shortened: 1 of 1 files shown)


DylanVann:
Thanks for the contribution! Looks like these changes have already been implemented though so I'm going to close this PR.

PR #277, opened 2018-09-06
missing break.

Yria:
when used with web caching strategy, it forces to enable cacheonly policy, which makes it impossible to use the option.


DylanVann:
That is not very good. Thank you!

PR #282, opened 2018-09-12
Add cache enum types

EQuimper:
Awesome library, I see in the readme the cache enum but typescript was complaint cause this is not existing in the type definition, so I add it :)


DylanVann:
This looks great, thank you!

PR #272, opened 2018-08-27
Fixes crash where activity was already destroyed

patrickkempff:
We saw some strange crashes in our brownfield app. We use a recycle list view with react-native-fast-image powered thumbnails. When the user navigates quickly to a new page and presses the back button really quick the app crashes with the following error:

```
Fatal Exception: java.lang.RuntimeException
java.lang.IllegalArgumentException: You cannot start a load for a destroyed activity
```
One of the threads in the glide repro: https://github.com/bumptech/glide/issues/1484

This PR basically adds some guards to the fast-image implementation. We have tested this in production (an app with over 2 million downloads and thousands of new users per day) without crashes. We did not notice any side effects.

**EDIT: i have pushed an new PR based on this one, that fixes the handling of uri's from require() on Android production builds, see #283**

Issue #246, opened 2018-07-12
Display a image through require.


giacomocerquone:
Hi all, I don't understand why this doesn't work

```
<FastImage
        source={{
          uri: require('../myfolder/myimage.jpg'),
          priority: FastImage.priority.high,
        }}
        fallback
        style={{
 width: 40, height: 40, borderRadius: 30, padding: 10,
}}
        resizeMode={FastImage.resizeMode.cover}
      />
```

If I use fallback true, and a default Image component is used, why I still must use a file url?
The error I get, of course, is this:

`undefined is not a function evaluating source.uri.startsWith('file://')`

I know how I can get around the problem (I'd just create a component that shows the default Image component or the fast image one based on certain condition), but I'd like to know if there is a proper way of doing this without the need of adding boilerplate code!

If it's normal that it doesn't work, it shouldn't be harder at all to implement it since you're already falling back to the Image default component that accepts a required local image

Mikunj:
```
<FastImage
source={require('../myfolder/myimage.jpg')}
/>
```
Have you tried doing that?

giacomocerquone:
jeez this works... so why @DylanVann wrote this in #160 :
> I am going to add support for local images. Probably soon / in the next major version. For now it does fallback to Image though.

Doing like you wrote it already works!
Anyway the api of this module are somehow awkward and the docs too.
What I mean with this is that the "source" prop can be both a required Image or an object defined by this lib specified in the doc that has this shape:

```
    source={{
      uri: 'https://unsplash.it/400/400?image=1',
      headers:{ Authorization: 'someAuthToken' },
      priority: FastImage.priority.normal,
    }}
```
and this is really confusing.

I propose two things:
- 1 I believe that the docs should be fixed ASAP (pr is coming) to specify that the source prop can be also a required image.

And in the near future:

- 2a There should be some prop like "fallbackSource" where you define the image to show when the lib goes in default mode (means when it uses the Image default component)

OR

- 2b The source prop becomes a string and should be used only for the uri (so it can be used interchangeably with required images, file url and remote url) and the other options like headers and priorities become other different props


Intuitively this is needed in order to give a clear shape to the source prop

gugakatsi:
@giacomocerquone just type there asset folder url for your image, after bundling it would be {{ uri: 'imagE_myimage'}} , and it will fallback to local image and u will't have to write boilerplate

n1ru4l:
On the current `master` the require/import is no longer working.

[...]

PR #283, opened 2018-09-14
[Android] Fixed local images in production builds (Fixes #273, #296 & #246)

### LINKED ISSUE #273: Can not show local Image
I have a problem When I user <Image />
 Somes times I load an local Image eg: <Image source={require('../images/Icon_thunder.png')}/> ,So I had Substitute <FastImage /> for <Image />
But Recently，Our online version still has many problems that cannot be loaded with local images.

patrickkempff:
This PR fixes the passing of require in production builds.

Context:
When using the following component: `<FastImage source={require('./image.png')} />`. The image loads just fine when running the android debug build. But when running the production variant, the same component is broken. This is because in debug mode the result of resolveAssetSource
is a `http://`resource as it want to fetch the image from the packager. In production a local drawable name is returned so it can be loaded from the Drawable folder in the APK itself. At the moment FastImage does not handle this case correctly.

Fix:
Refactored the mechanism that resolves the resource uri that Glide can understand. I investigated how the stock RN Image component handles this case, as it turns out `ImageSource` is used to transform the output of resolveAssetSource to a local `res:/` uri. We then take this uri and transform to a format that Glide understands (android.resource://). This will be handle by `FastImageSource` which is build on top of `ImageSource`

This PR is build on top of #272

Done:
- [x] Support require() in production mode (Fixes #273, #296 & #246)
- [x] Fix crash when passing null uri
- [x] Prevents crash when activity is already destroyed
- [x] Fixes possible breakage with newer RN version by using `Image.resolveAssetSource` instead of importing it directly as it is not offically supported.
- [x] Make sure `data:image/png;base64` uri's still work.
- [x] Added missing breaks in switch statement for caching strategy, thanks @Yria! (#282)


manhhoan73:
Can you give me the example? I build the release not display the image.

fierysolid:
@patrickkempff does this add support for bundled images? Documentation on bundled images can be found [here](https://facebook.github.io/react-native/docs/images#images-from-hybrid-app-s-resources)

patrickkempff:
@fierysolid not at the moment, I am working on a solution to support this.

> @patrickkempff does this add support for bundled images? Documentation on bundled images can be found [here](https://facebook.github.io/react-native/docs/images#images-from-hybrid-app-s-resources)

patrickkempff:
@manhhoan73 can you tell me the steps you took integrating this PR?

DylanVann:
screen shot 2018-10-22 at 10 59 15 pm

😢

src/index.js
```diff
@@ -34,7 +32,7 @@ class FastImage extends Component {
             ...props
         } = this.props
 
-        const resolvedSource = resolveAssetSource(source)
+        const resolvedSource = Image.resolveAssetSource(source)
```
DylanVann: Awesome. It was not a nice looking import before.

android/src/main/java/com/dylanvann/fastimage/FastImageViewManager.java
```diff
@@ -153,4 +182,50 @@ public float getGranularityPercentage() {
...
+    private static boolean isAcitityDestroyed (Activity activity) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN_MR1) {
+            return activity.isDestroyed() || activity.isFinishing();
+        } else {
+            return activity.isFinishing() || activity.isChangingConfigurations();
+        }
+
+    }
+
+    private void warnImageSource(Context context, String uri) {
```
DylanVann: Doesn't seem to be used.

DylanVann:
I resolved some of the conflicts and merged this in another PR (https://github.com/DylanVann/react-native-fast-image/pull/324).

Need to do some more testing before another release, but it looks good. Thank you!

PR #287, opened 2018-09-17
allow overwriting imageContainer styles

### LINKED ISSUE #286: allow overwriting imageContainer styles (aka allow disabling overflow: 'hidden' style for shadow)
This style:
https://github.com/DylanVann/react-native-fast-image/blob/47965fa952a1bcd480c2b3e5e5730cdf3248f89a/src/index.js#L81

Cannot be overwritten because of the order the styles are applied https://github.com/DylanVann/react-native-fast-image/blob/47965fa952a1bcd480c2b3e5e5730cdf3248f89a/src/index.js#L43

When using shadows, eg:

```js

<FastImage
  style={{ shadowOpacity: 0.2,
    shadowColor: themeStyles.colors.black,
    shadowRadius: 2,
    shadowOffset: {
      height: 1,
      width: 1,
    },
    height: 20,
    width: 20,
  }}
/>

```

The text shadow will always be cropped.

If we would change the order of the styles being applied, users could overwrite the `overflow: 'hidden'` style.

@DylanVann I will send a pull request that will fix this issue soon!

n1ru4l:
Closes #286

### DIFF
--- a/src/__snapshots__/index.test.js.snap
+++ b/src/__snapshots__/index.test.js.snap
@@ -5,11 +5,11 @@ exports[`FastImage renders correctly. 1`] = `
   style={
     Array [
       Object {
-        "height": 44,
-        "width": 44,
+        "overflow": "hidden",
       },
       Object {
-        "overflow": "hidden",
+        "height": 44,
+        "width": 44,
       },
     ]
   }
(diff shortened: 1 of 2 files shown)


DylanVann:
Yeah this seems more reasonable. Thank you!

PR #303, opened 2018-10-05
Fixed unordered setting props issue

yamade:
## Background

I had an issue sometimes `_onFastImageXXX` callback doesn't get called in iOS.

The root cause is the props on iOS side are stored in an `NSDictionary`, the keys are unordered.
This means props have no guaranteed order in which they're updated.

https://github.com/facebook/react-native/issues/8181

Fix for #304

## Changes
Use `didSetProps` and introduce `reloadImage` to make sure callback gets call after set props


DylanVann:
Awesome! We probably actually want this for a bunch of props on iOS.

Issue #306, opened 2018-10-09
Could not find method implementation()


oskaryil:
React Native version: `0.56.0`
react-native-fast-image version: `5.0.10`

When running `react-native run-android` I'm getting a build error:
```
* Where:
Build file '/Users/oskar/Documents/Coding/SHL/app/node_modules/react-native-fast-image/android/build.gradle' line: 55

* What went wrong:
A problem occurred evaluating project ':react-native-fast-image'.
> Could not find method implementation() for arguments [com.facebook.react:react-native:+] on object of type org.gradle.api.internal.artifacts.dsl.dependencies.DefaultDependencyHandler.
```
Any idea on how to solve this?

Lizhooh:
me too.

DISKONEKTeD:
i am having the same issue

PR #307, opened 2018-10-09
Changed implementation to compile to support new Android build tools

oskaryil:
Fixes: #306


patrickkempff:
I am not sure this is the right fix. Compile is deprecated and will be removed in favour of implementation. I think you should update your gradle version.

oskaryil:
@patrickkempff Yeah you're right. I figured that out just right after posting this really haha.

PR #324, opened 2018-10-23
Fix local images in production builds.

DylanVann:


yeomann:
hello, this PR seems merged but is this released on npm so that we can start using the fix for production builds

nafisholeh:
not released yet
see this https://github.com/DylanVann/react-native-fast-image/blob/master/CHANGELOG.md

PR #328, opened 2018-10-26
Improve error message when bundled resource could not found.

patrickkempff:
A team member of mine spend a couple of hours debugging this error:
`Couldn't find com.facebook.react.uimanager.annotations.ReactProp.source`

There was no clear indicator of what the underlaying problem was.  As it turns out the error was caused by missing assets. To make sure other developers don't waste the same amount of time as we did, we improved the error message.


patrickkempff:
@DylanVann awesome!

PR #336, opened 2018-11-01
Fix TS typings

kidcosmic:
Fixes error: `node_modules/react-native-fast-image/src/index.d.ts:129:30 - error TS2694: Namespace 'FastImage' has no exported member 'cache'.`


DylanVann:
This has been fixed on master. New release will be coming soon.

PR #350, opened 2018-11-23
Update FastImageViewManager.java

ghost:
FIX:failed to load image while ui rebuilding


DylanVann:
This would need more explanation to be actionable. Please open another PR if you think it's still relevant. Thank you!

PR #378, opened 2018-12-25
Added Palette support, placeholder color, fadeIn boolean property

eladgel:
### Added color palette support
Passing to onLoad function through the event object all the available Swatches with their values **rgba**, **hsl**, **titleTextColor**, **bodyTextColor**
according to Google's docs

### Added two more properties to image loading, 
1. The user is now able to load image with a fade in transition using the **fadeIn** property
2. The user is now able to pass **placeholderColor** #hex as a property


eladgel:
Echo

android/src/main/java/com/dylanvann/fastimage/FastImageRequestListener.java
```diff
@@ -23,13 +28,64 @@
...
     }
 
+
+    private static WritableMap mapFromResourceWithPalette(Drawable resource, Palette palette) {
+        WritableNativeMap swatchMap = new WritableNativeMap();
+
+        addHexStringIfSwatchValid(palette.getDominantSwatch(), swatchMap, "dominantSwatch");
+        addHexStringIfSwatchValid(palette.getLightVibrantSwatch(), swatchMap, "lightVibrantSwatch");
+        addHexStringIfSwatchValid(palette.getVibrantSwatch(), swatchMap, "vibrantSwatch");
+        addHexStringIfSwatchValid(palette.getDarkVibrantSwatch(), swatchMap, "darkVibrantSwatch");
```
DylanVann: I'm not sure what most of this is trying to achieve, what does "palette support" mean?

Ilphrin: I think it is about this: https://developer.android.com/reference/android/support/v7/graphics/Palette

DylanVann:
I don't think this functionality needs to go in this library.

I would be interested in a `fade` property that made images use the normal fade animation on Android and on iOS the same sort of animation. Thank you!

luanmm:
> I don't think this functionality needs to go in this library.
> 
> I would be interested in a `fade` property that made images use the normal fade animation on Android and on iOS the same sort of animation. Thank you!

I agree, but... (to give another point of view regarding this subject)... it should be a way to get a reference to the loaded image if this functionality should be somewhere else (like a third-party library or custom code). I tried to find something related to it, but it seems that today is not possible to get anything about the loaded resource.

That being said, this contribution seems to be useful for some use cases (like defining dynamic backgrounds from image colors, which is what I'm looking for).

PR #386, opened 2019-01-03
Support WebP images on iOS 12

### LINKED ISSUE #298: WebP not showing on iOS 12
"react": "16.5.0",
"react-native": "0.57.1",
"react-native-fast-image": "^5.0.10"

### LINKED ISSUE #385: Animated WebP display issues (With "Fix")
Currently there is a problem with displaying animated webp without transparency on iOS12 devices.
This problem was fixed in SDWebImage v4.4.1
Just from skipping through the SDWebImage changelogs there doesnt appear to be breaking changes on first sight so it would be nice if the version could be updated

IljaDaderko:
Should
fix #385
fix #298

I tested this in Debug and Release builds with my own app (ios 12) via local images i.e.
```js
<FastImage source={require("../assets/myImage.webp")} />
```

Seems to work without any trouble. Since we are aiming for fast image lib, I think `SDWebImage/WebP` is a good candidate to be included with CocoaPod file?

### DIFF
--- a/react-native-fast-image.podspec
+++ b/react-native-fast-image.podspec
@@ -22,5 +22,6 @@ Pod::Spec.new do |s|
   s.dependency 'React'
   s.dependency 'SDWebImage/Core'
   s.dependency 'SDWebImage/GIF'
+  s.dependency 'SDWebImage/WebP'
   s.dependency 'FLAnimatedImage'
 end


IljaDaderko:
Hmm it still does not display webp images without alpha layer for some reason, i.e. jpegs that were converted to WebP.

I think this is related PR https://github.com/DylanVann/react-native-fast-image/pull/387

@ds8k Judging by `Podfile.lock`

```
  - SDWebImage/Core (4.4.3)
  - SDWebImage/GIF (4.4.3):
    - FLAnimatedImage (~> 1.0)
    - SDWebImage/Core
  - SDWebImage/WebP (4.4.3):
    - libwebp (~> 0.5)
```

SDWebImage is already using version `4.4.3`, but I'm not sure why this project has it as submodule.

jjgp:
Would this support those who use `react-native link` instead of using CocoaPods to integrate third party? Seems there's an additional step to do so

IljaDaderko:
@jjgp I'm not sure but I think https://github.com/DylanVann/react-native-fast-image/pull/386#issuecomment-451239433 addresses `react-native link` part? I only tested CocoaPods implementation however.

jjgp:
I believe you're correct I will follow up there!

jjgp:
@IljaDaderko I'm not a maintainer, however, am eager to get this fix as well. I don't see a problem with them being separate PRs. They should be merged into the same release. Only comment I have is that we may want to assert (or at least test for now) that the pod version of SDWebImage is the same(ish) version as the vendor submodule in the other PR.

DylanVann:
It's merged! Sorry, there was another PR open for this and I merged that one. Thank you though @IljaDaderko !

https://github.com/DylanVann/react-native-fast-image/releases/tag/v5.1.3

Issue #371, opened 2018-12-14
Memory Leak on iOS: FastImage not released from memory when callbacks onLoad is set


StevenMasini:
An engineer at my company noticed a memory leak on iOS when the props `onLoad` is set on iOS.
FastImage doesn't get released when it should be.

After investigating the source code, I found out that using a `strong` reference to `self` to call `sendOnLoad ` in `sd_setImageWithURL`.

I suspect this line to be the cause of this memory leaks.

https://github.com/DylanVann/react-native-fast-image/blob/141a4a02931ba9e02ee169ba4543e77d924b6a79/ios/FastImage/FFFastImageView.m#L167

I can submit a Pull Request to transform this `strong` reference into a `weak` reference, which should fix the memory leak.

oblador:
Sounds like a good idea to me 🤷‍♂️

StevenMasini:
Ok sorry for the delay, I did the fix on my fork not long after opening this issue, but I forgot to submit the PR. Now it's done I submitted the PR.

marf:
Hello @StevenMasini, I have tried your PR but it does not compile. (as I reported you in the PR)
What should I do?

Thank you,
Marco

StevenMasini:
@marf Hey sorry for the delay. I am working on it since yesterday. I found a very fast and easy to reproduce memory leaks when you set the `onError` props in `<FastImage>`, but it's not really obvious yet what's causing this memory leaks.

Also I am sorry for the fix I submitted earlier, it's been a while that I haven't done Objective C, so I forgot that `typeof(self)` will actually define it as a pointer already. So the correct syntax for the fix I submitted is `__weak typeof(self) weakSelf = self;` without the `*`.

But still that's not enough to fix the memory leak. So I will keep investigating.

truongluong1314520:
HI @StevenMasini , if the `onError` is not be set, there is no memory leak right?

StevenMasini:
@truongluong1314520 You can set the `onError` but just don't try to get any parameters from it. 

`onError` don't pass any parameters. 

For instance do 
```js
onError={() => {
// your implemention
}}
```
but don't 
```js
onError={error => {
// this will create a memory leak, the same for every callback, just strictly follow the documentation
console.log(error)
}}
```

StevenMasini:
I made a new Pull Request here https://github.com/DylanVann/react-native-fast-image/pull/433

StevenMasini:
[...]

PR #398, opened 2019-01-29
Memory Leak on iOS: FastImage not released from memory when callbacks onLoad is set

StevenMasini:
# Changelog

Solution to the open issue #371

- Use a `weak` reference to `self` to call `sendOnLoad` in the `sd_setImageWithPreviousCachedImageWithURL:placeholderImage:options:progress:completed:`

Holding a `strong` reference to `self` in a block can lead to memory leaks as the block will hold the reference until it is called.

[Always pass weak reference of self into block in ARC?](https://stackoverflow.com/a/20032131)


adammcarth:
I seem to be getting two compile errors in XCode when running these changes:

1) `Implicit conversion of an Objective-C pointer to 'FFFastImageView *const __weak *' is disallowed with ARC`.
```objectivec
__weak typeof(self) *weakSelf = self;
```

2) `Bad receiver type 'FFFastImageView *const __weak *'`.
```objectivec
[weakSelf sendOnLoad:image];
```

StevenMasini:
@adammcarth Okay I will have a look at it. Sound weird tho, you are not supposed to pass a strong pointer into a block.

marf:
> I seem to be getting two compile errors in XCode when running these changes:
> 
> 1. `Implicit conversion of an Objective-C pointer to 'FFFastImageView *const __weak *' is disallowed with ARC`.
> 
> ```objc
> __weak typeof(self) *weakSelf = self;
> ```
> 
> 1. `Bad receiver type 'FFFastImageView *const __weak *'`.
> 
> ```objc
> [weakSelf sendOnLoad:image];
> ```

Hello @adammcarth I have the same problem trying to applying the fix.
What should we do?

Thank you,
Marco

StevenMasini:
Yeah sorry guys my bad, the correct syntax is `__weak typeof(self) weakSelf = self;`. It's been a while since I wrote Objective C, I forgot that `typeof(self)` will already define the type as a pointer.

StevenMasini:
Well okay I am stupid, I have been chasing a ghost 😓 

Actually if you implement `onLoadStart `, `onError` or `onLoadEnd` and you try to retrieve a parameter from it. Then you will end up with a huge memory leak.

I was trying to retrieve `error` from `onError`, but actually the native part doesn't return it at all.

Well at least I have discovered that you can't retrieve a parameter that doesn't exist from a callback.

Anyway we still need to give a weak pointer of self to the completion block of SDWebImage.

marf:
Hello, @StevenMasini and thank you.
So has this fix fixed the memory leaks?

Also changing from RTCBubblingEventBlock to RTCDirectEventBlock fixes the fact of the parameters to the callback which may generate memory leaks?

Thank you,
MArco

StevenMasini:
I am preparing another Pull Request with more fixes. I am closing this Pull Request as the new Pull Request will contain these fix and more.

Issue #285, opened 2018-09-16
support for resizeMethod


mrafei:
Hi,

I have a list of rather large images, and I want smooth scroll behavior. If I use react native's `Image` with `resizeMethod` set to 'resize`, I get a performance boost over `FastImage` but I lose the caching and loading optimizations of `FastImage`.
Is there a reason why `FastImage` doesn't support `resizeMethod` prop?

thanks.

sesam:
@mrafei are you loading big images, but showing them in a smaller format? Could you load the smaller images first, and only load the big image when needed (like maybe after someone presses/clicks the small image)

n1ru4l:
@mrafei Have you checked the https://github.com/DylanVann/react-native-fast-image#resizemode-enum property?

adammcarth:
Yeah we're having major issues with scrolling performance on Android without the Android-specific `resizeMethod` property being set to `"resize"`.

adammcarth:
Alright, for anyone else having this issue - I'd suggest making sure you're running at least version [`5.1.0`](https://github.com/DylanVann/react-native-fast-image/releases/tag/v5.1.0). We were running version `4.0.14`, and the upgrade alone seemed to dramatically improve performance from _unusable_ to acceptable. This is (I suspect) largely due to the upgraded major version of [Glide](https://bumptech.github.io/glide/).

For those still experiencing problems on Android with large down-sized images, I've [created a fork of the latest version](https://github.com/adammcarth/react-native-fast-image#resizeimageandroid-object-android-only) which introduces a new prop to manually downsize images before they're loaded into the RAM. The only caveat is that you must know the exact resolution that you want to downsize to.

Usage Example:

```jsx
<View style={{ width: 450, height: 275 }}>
  <FastImage
    source={{ uri: 'http://example.com/huge-image-3600x2200.png' }}
    resizeMode="contain"
    resizeImageAndroid={{ width: 450, height: 275 }} // <-------
  />
</View>
```

I'll submit this as a PR too, no guarantees if it'll be accepted though.

mralj:
I haven't tried forked version, but I have allied  "dirty fix" in `FastImageViewManager.java`
Just for test I have changed this line 
`.apply(FastImageViewConverter.getOptions(source))` to
`.apply(FastImageViewConverter.getOptions(source).override(600, 400) )`

[...]

PR #401, opened 2019-01-31
Add Option To Resize Images On Android

adammcarth:
As discussed in issue #285, there are some cases where very large, down-sampled images can create performance issues on Android due to the entire image being loaded in the device's RAM.

There is an option in React Native's default `<Image />` component which allows you to specify [`resizeMethod="resize"`](https://facebook.github.io/react-native/docs/image#resizemethod). This uses Facebook's [Fresco](https://frescolib.org/) library to automatically downsample the image itself before being loaded into the devices memory, but since `react-native-fast-image` uses Glide - there appears to be nothing similar available.

The solution I came up with was to offer the ability to manually down sample the image size before it's loaded into the memory using a new prop called `resizeImageAndroid`. It uses Glide's [`.override()`](https://bumptech.github.io/glide/javadocs/430/com/bumptech/glide/request/RequestOptions.html#override-int-int-) method to achieve this.

The only caveat is that you must know the exact image size, but it's nice to at least have this option if it's critical for the performance of large images on Android.

Example Usage:

```jsx
<View style={{ width: 450, height: 275 }}>
  <FastImage
    source={{ uri: 'http://example.com/huge-image-3600x2200.png' }}
    resizeMode="contain"
    resizeImageAndroid={{ width: 450, height: 275 }} // <-------
  />
</View>
```

### DIFF
--- a/src/index.js.flow
+++ b/src/index.js.flow
@@ -24,7 +24,7 @@ export type ResizeMode = $ReadOnly<{|
     center: 'center',
 |}>
 
-export type ImageResizeAndroid = $ReadOnly<{|
+export type ResizeImageAndroid = $ReadOnly<{|
     width: number,
     height: number,
 |}>
@@ -64,7 +64,7 @@ export type FastImageProps = $ReadOnly<{|
     source: FastImageSource | number,
 
     resizeMode?: ?ResizeModes,
-    imageResizeAndroid?: ?ImageResizeAndroid,
+    resizeImageAndroid?: ?ResizeImageAndroid,
     fallback?: ?boolean,
     testID?: ?string,
 |}>


android/src/main/java/com/dylanvann/fastimage/FastImageViewManager.java
```diff
@@ -114,6 +117,13 @@ public void setResizeMode(FastImageViewWithUrl view, String resizeMode) {
         view.setScaleType(scaleType);
     }
 
+    @ReactProp(name = "resizeImageAndroid")
+    public void setImageResize(FastImageViewWithUrl view, ReadableMap imageSizeOverride) {
+        // Re-run Glide with width and height override values set.
```
badeleux: Wouldn't it be better to put `imageResize` into source, so we don't have to rerun Glide? It would be beneficial for iOS as well.

madandrija:
Any way I can help with getting this PR merged in? 

We are currently maintaining a fork with this solution implemented in the latest FastImage release and I'd prefer if I could just use the original and drop the fork 😁

DylanVann:
I believe Glide does downsampling for us without any extra config.

> Glide automatically limits the size of the image it holds in cache and memory to the ImageView dimensions.

From https://futurestud.io/tutorials/glide-image-resizing-scaling

christopherklint97:
I am also doing as @madandrija states by maintaining a forked version with this pr implemented.

I am not a expert with Glide but I did notice a huge difference when I implemented this solution (pretty much night and day). Really think it would be helpful for other android users to have at their disposal.

DylanVann:
You may be getting better performance because the quality of your rendered images is less if you are not accounting for DPI.

Assuming Glide's auto downsampling works that's the reason I would guess you're seeing better performance.

We could try to create a reproduction of this issue to check if that's the case, or if the auto sampling based on the view size is not working for some reason.

madandrija:
> I am also doing as @madandrija states by maintaining a forked version with this pr implemented.
> 
> I am not a expert with Glide but I did notice a huge difference when I implemented this solution (pretty much night and day). Really think it would be helpful for other android users to have at their disposal.

We've updated our fork to the latest version and still we need the fork in order to speed up image loading (e.g. when scrolling through a FlatList, w/o the fork there were sometimes "blanks" until images loaded, whereas with the fork, everything was running smoothly), and I have [two](https://mdbrzallpics.azureedge.net/home-transformationBE763C5020688A49B6480B8C76CA0594.jpg) [examples](https://mdbrzallpics.azureedge.net/body-killerEBD3C6DAE6F32E4E88ED8FCFE502E469.jpg) of images used.

Regarding repro, still thinking about how to set up a minimal example, either on GitHub or in Snack.

PR #406, opened 2019-02-14
preload cache control

michalmelich-gl:
added a parameter to preload method to control cache options on iOS


DylanVann:
I think this isn't needed because `cacheControl` is set per source. So if you want to preload a bunch of images with a certain value just make sure the source objects have it.

DylanVann:
On second look, it looks like that doesn't work on iOS because SDWebImage didn't support setting the `cacheControl`/equivalent per image being preloaded, maybe it does now.

Issue #432, opened 2019-03-20
Problems with iOS Image download failure


StevenMasini:
Hi @DylanVann 

I am addressing two problems here, but with the same root cause.

Fortunately I already know how to fix these issues on iOS. 
Also I checked on Android and it seems that Android auto-retry when images failed to download. 

So as far as I know Android isn't affected by this problem.

### Problem 1:

Image failed to download, they won't auto-retry the download and no error will be broadcasted.
The method `onError` is triggered but won't return the error. 

*I understand that it's difficult to unify Android and iOS error handling*

### Problem 2:

I explained on this thread #394

Let's say we have **ImageCanvas1** on **Page 1** who failed to download the image, then I push **Page 2** with **ImageCanvas2** who successfully download the picture.

When I pop back to **Page 1,** **"ImageCanvas1"** isn't aware of the download by **ImageCanvas2** and is still empty.

### Solution

1. Apparently we can't only rely on the flag `SDWebImageRetryFailed` to auto-retry when an image download failed. We need to implement our own auto-retry mechanism. https://github.com/SDWebImage/SDWebImage/issues/2587

2. We can post a notification via `NSNotificationCenter` to all the instance of `FFFastImageView` sharing the same `source.url` to let them know that the resource has been downloaded successfully by one of them. 

I am not sure that `SDWebImageRetryFailed ` will manage all failure scenario but for the second solution I tested it and it works perfectly.

antitoine:
For anyone looking for a quick solution for working around using only React Native : https://github.com/DylanVann/react-native-fast-image/issues/394#issuecomment-566664741

Issue #421, opened 2019-02-26
Crash due to memory issue iOS & Android


marf:
Hello,
I am heavily using Fast Image in my app since I have large lists (with 300+ elements) to render with an image on the every item.

I have noticed that on old devices for example iPhone 6, 5s etc. and a lot of Android phones the app crashes and in Xcode (for iPhones) the crash reported: Memory issue

If I remove fast image and use normal Image component the crash dissapear.

What should I do?
Is this due to a memory leak?

I have tried to follow this #371 which reports how to fix memory leaks but it does not work.

What should I do?

Best regards,
Marco

StevenMasini:
Hi @marf 

My previous PR was only for iOS, I don't know if there is memory leaks on Android, and if there is I don't know what cause them neither.

Can I ask you to try to see if my new PR #433 at least fix the memory leak on iOS for you ?

I remember you said you couldn't compile #371 before, and I found out that there is a bug with npm & yarn. I put instruction about how to try my PR in the comment.

Please try it out and let me know.

evanjmg:
I'm think still getting crashes due to memory? The crashes are very few and these devices have less than 5% memory left on device. 

```
Fatal Exception: NSInvalidArgumentException-[UIViewAnimationState url]: unrecognized selector sent to instance 0x110c428e0 Raw Text
--
0 | CoreFoundation | __exceptionPreprocess
4 | CoreFoundation | _CF_forwarding_prep_0
5 | FFFastImageView.m line 79-[FFFastImageView imageDidLoadObserver:]
```

StevenMasini:
@evanjmg 

```
Fatal Exception: NSInvalidArgumentException-[UIViewAnimationState url]: unrecognized selector sent to instance 0x110c428e0 Raw Text
```

This isn't related to memory leak error. Most likely what you said here https://github.com/DylanVann/react-native-fast-image/pull/433#discussion_r272848702

ammoradi:
any news about this?
in **Android**, using this lib causing huge memory usage and crashes on low-android-versions devices!

afeichuanqi:
> 关于这个的任何消息？
> 在**安卓中**，使用此lib中会导致大量内存使用并在低版本的Android的设备上崩溃！

is there any news

Mihai-github:
Hi guys, 

[...]

Issue #394, opened 2019-01-12
iOS - Caching Image failures


KevinColemanInc:
I am loading 20 images hosted in the USA on a very slow wifi connection from Asia. In the browser, the images fetch, but its slow.

From what I can tell FastImage if the image loading fails (`onError` is triggered), it seems to cache the error instead of trying again later. Only deleting the app and re-launching it again will FastImage be able to fetch the images again.

I don't know iOS / Swift well enough to confirm my hypothesis, but is anyone else seeing this?

StevenMasini:
@KevinColemanInc I think I am experiencing the same issue as you. 

You are indeed right, I investigated the issue and right now on iOS when the "ImageCanvas" (I will call it like this) fail to download an image. Well first it fail silently, the error isn't being broadcast to the RN side, and second it won't try to download the image again.

So my issue is a bit different than yours. I notice this because in my app I have the same picture being display on several screen, and sometime it loads in some screen "ImageCanvas" but not in others.

e.g I have "ImageCanvas1" on page 1 who failed to download the image, then I push page 2 with "ImageCanvas2" who load successfully the picture. When I pop back to page 1 "ImageCanvas1" still haven't load the picture.

So I have fixed my issue on my own branch but haven't submit a PR for it yet. Also it's an iOS related issue, this doesn't reproduce on Android as Glide will automatically re-trigger the download if it fails.

But as I said your issue is different, and I should made a fix that handle all the scenario.

I will open a new PR for this, and will keep you updated about it.

StevenMasini:
@KevinColemanInc I made a Pull Request that should fix your issue https://github.com/DylanVann/react-native-fast-image/pull/433

StevenMasini:
@KevinColemanInc I think you can try my PR https://github.com/DylanVann/react-native-fast-image/pull/433 to see if it fix your problem. Let me know if you have feedback about it.

KevinColemanInc:
@StevenMasini I pushed your fix out to production and I haven't seen it crash yet, but I also have not explicitly tested for this case.

StevenMasini:
@KevinColemanInc Thank you for helping me out.

The PR include an auto-retry logic, so this should solve your problem. Let me know if you figured out anything else or if you have further feedbacks. 👍

StevenMasini:
@KevinColemanInc Any updates ? Do you still have the crash ?

evanjmg:
[...]

PR #433, opened 2019-03-20
iOS: Memory Leak, Auto-retry download, Shared Resource Propagation

StevenMasini:
# Changelogs

**1- Use weak self reference in progression and completion block to avoid memory leaks**

It's recommended to never retain the strong reference to self into blocks.

*Blocks maintain strong references to any captured objects, including self, which means that it’s easy to end up with a strong reference cycle*

[Avoid Strong Reference Cycles when Capturing self](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/WorkingwithBlocks/WorkingwithBlocks.html#//apple_ref/doc/uid/TP40011210-CH8-SW16)

**2- Broadcast a notification when different instance share the same source succeed to download an image**

As mentioned in #432, when two instance of **FFFastImageView** share the same source, but one of them fail to download while the second succeed, the result of the second won't be shared with the first one. Which result with instance of **FFFastImageView** staying empty.

To remedy this issue I used `NSNotificationCenter` to attach an observer and post a notification whenever an instance of **FFFastImageView** succeed to download. Works very well on my own project, it's a game changer for me.

**3- Implement auto-retry failed download logic**

As mentioned in #432, we can't only rely on `SDWebImageRetryFailed` to auto-retry when a download failed.

Also apparently this feature has been requested on the SDWebImage repository and is still waiting for implementation.
https://github.com/SDWebImage/SDWebImage/issues/2587

So I implemented my own auto-retry mechanism which will retry after 1.5s, 3s and 4.5s before failing. In total it makes the timeout for the image download to 9s.

Let me know if you want to bring more flexibility to this feature.

**4- Update Objective-C syntax to highlight instance properties**

Being a former Objective-C iOS engineer myself, I updated the code to the modern Objective-C syntax.

[Adopting Modern Objective-C](https://developer.apple.com/library/archive/releasenotes/ObjectiveC/ModernizationObjC/AdoptingModernObjective-C/AdoptingModernObjective-C.html) 
[Migrating to Modern Objective-C](https://download.developer.apple.com/wwdc_2012/wwdc_2012_session_pdfs/session_413__migrating_to_modern_objectivec.pdf) See slides 47 and 48

Updating the syntax helped me to realized that their were more invocation to strong self in blocks that I initially thought. See #398 

**5- Use RCTDirectEventBlock instead of RCTBubblingEventBlock**

From what I understand from the React Native documentation.

[RCTBubblingEventBlock](https://facebook.github.io/react-native/docs/native-components-ios#events)  are used for UI related events (e.g: the user pinch the map)
[RCTDirectEventBlock](https://facebook.github.io/react-native/docs/custom-webview-ios#adding-new-events-and-properties) are used for more straight events (e.g web page failed to load callbacks).

So I replaced all the **RCTBubblingEventBlock** types by **RCTDirectEventBlock** as we don't handle user interaction.

Also found this [gist](
https://gist.github.com/chourobin/f83f3b3a6fd2053fad29fff69524f91c#ios) about the subject.

### DIFF
--- a/ios/FastImage/FFFastImageView.m
+++ b/ios/FastImage/FFFastImageView.m
@@ -119,8 +119,7 @@
         }];
         
         // Set priority.
-        SDWebImageOptions options = 0;
-        options |= SDWebImageRetryFailed;
+        SDWebImageOptions options = SDWebImageRetryFailed; // Auto-retry to download if failed
         switch (_source.priority) {
             case FFFPriorityLow:
                 options |= SDWebImageLowPriority;
--- a/ios/FastImage/FFFastImageView.m
+++ b/ios/FastImage/FFFastImageView.m
@@ -31,14 +31,14 @@
     }
 }
 
-- (void)setOnFastImageLoadEnd:(RCTBubblingEventBlock)onFastImageLoadEnd {
+- (void)setOnFastImageLoadEnd:(RCTDirectEventBlock)onFastImageLoadEnd {
     _onFastImageLoadEnd = onFastImageLoadEnd;
     if (self.hasCompleted) {
         _onFastImageLoadEnd(@{});
     }
 }
 
-- (void)setOnFastImageLoad:(RCTBubblingEventBlock)onFastImageLoad {
+- (void)setOnFastImageLoad:(RCTDirectEventBlock)onFastImageLoad {
     _onFastImageLoad = onFastImageLoad;
     if (self.hasCompleted) {
         _onFastImageLoad(self.onLoadEvent);
@@ -52,7 +52,7 @@
     }
 }
 
-- (void)setOnFastImageLoadStart:(RCTBubblingEventBlock)onFastImageLoadStart {
+- (void)setOnFastImageLoadStart:(RCTDirectEventBlock)onFastImageLoadStart {
     if (_source && !self.hasSentOnLoadStart) {
         _onFastImageLoadStart = onFastImageLoadStart;
         onFastImageLoadStart(@{});
--- a/ios/FastImage/FFFastImageView.m
+++ b/ios/FastImage/FFFastImageView.m
@@ -1,11 +1,17 @@
 #import "FFFastImageView.h"
 
-@implementation FFFastImageView {
-    BOOL hasSentOnLoadStart;
-    BOOL hasCompleted;
-    BOOL hasErrored;
-    NSDictionary* onLoadEvent;
-}
+
+@interface FFFastImageView()
+
+@property (nonatomic, assign) BOOL hasSentOnLoadStart;
+@property (nonatomic, assign) BOOL hasCompleted;
+@property (nonatomic, assign) BOOL hasErrored;
+
+@property (nonatomic, strong) NSDictionary* onLoadEvent;
+
+@end
+
+@implementation FFFastImageView
 
 - (id) init {
     self = [super init];
@@ -27,43 +33,43 @@
 
 - (void)setOnFastImageLoadEnd:(RCTBubblingEventBlock)onFastImageLoadEnd {
     _onFastImageLoadEnd = onFastImageLoadEnd;
-    if (hasCompleted) {
+    if (self.hasCompleted) {
         _onFastImageLoadEnd(@{});
     }
 }
 
 - (void)setOnFastImageLoad:(RCTBubblingEventBlock)onFastImageLoad {
     _onFastImageLoad = onFastImageLoad;
-    if (hasCompleted) {
-        _onFastImageLoad(onLoadEvent);
+    if (self.hasCompleted) {
+        _onFastImageLoad(self.onLoadEvent);
     }
 }
 
 - (void)setOnFastImageError:(RCTDirectEventBlock)onFastImageError {
     _onFastImageError = onFastImageError;
-    if (hasErrored) {
+    if (self.hasErrored) {
         _onFastImageError(@{});
     }
 }
 
 - (void)setOnFastImageLoadStart:(RCTBubblingEventBlock)onFastImageLoadStart {
-    if (_source && !hasSentOnLoadStart) {
+    if (_source && !self.hasSentOnLoadStart) {
         _onFastImageLoadStart = onFastImageLoadStart;
         onFastImageLoadStart(@{});
-        hasSentOnLoadStart = YES;
+        self.hasSentOnLoadStart = YES;
     } else {
         _onFastImageLoadStart = onFastImageLoadStart;
-        hasSentOnLoadStart = NO;
+        self.hasSentOnLoadStart = NO;
     }
 }
 
 - (void)sendOnLoad:(UIImage *)image {
-    onLoadEvent = @{
-                    @"width":[NSNumber numberWithDouble:image.size.width],
-                    @"height":[NSNumber numberWithDouble:image.size.height]
-                    };
+    self.onLoadEvent = @{
+                         @"width":[NSNumber numberWithDouble:image.size.width],
+                         @"height":[NSNumber numberWithDouble:image.size.height]
+                         };
     if (_onFastImageLoad) {
-        _onFastImageLoad(onLoadEvent);
+        _onFastImageLoad(self.onLoadEvent);
     }
 }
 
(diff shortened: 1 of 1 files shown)


StevenMasini:
Just in case someone wants to try this branch on their own project.

I notice there is an issue with **npm** & **yarn** when you try to point a dependency to a direct commit which contain a submodule.

The git-submodules aren't loaded.

See issue https://github.com/yarnpkg/yarn/issues/1488 and https://github.com/npm/npm/issues/1876

So if in your `package.json`, you try to point to the latest commit of this PR, then the folder `node_module/react-native-fast-image/ios/Vendor/SDWebImage` will become empty, and you won't be able to build the project.

So in order for you to try this solution, you will need to

In your **package.json** add

```json
"react-native-fast-image": "StevenMasini/react-native-fast-image#27e9eacc0c9cca6fa16ec8d342a8d804b926c6df",
``` 


and in your **Podfile** (hopefully you use Cocoapods) add

```
pod 'SDWebImage'
```

evanjmg:
@DylanVann any feedback on this?

evanjmg:
I tried this out and I'm getting repeated images in cells. Is there anything changed in this that could cause this issue?

StevenMasini:
@evanjmg 

Do you means the warning that `<FlatList />` will trigger if you don't specify a `key` in the `renderItem` method ?

Then if that the case you just need to add a prop `key` to your element. 

Something like
```js
using.map( (item, index) => {
     <FastImage key={`0x${index}`} source={{ uri }} />
})
```

If not can you give me more detail ?

StevenMasini:
@DylanVann I am asking once again to review this PR 🙏 

I am already using it in Production on my own app via `@steven/react-native-fast-image` module.

Also this PR is a follow up and aim to fix this issues #371 #421 #394

evanjmg:
I'm using it in production too 👍

ios/FastImage/FFFastImageView.m
```diff
@@ -14,82 +20,95 @@ - (id) init {
...
+    self.onLoadEvent = @{
+                         @"width":[NSNumber numberWithDouble:image.size.width],
+                         @"height":[NSNumber numberWithDouble:image.size.height]
+                         };
+    if (self.onFastImageLoad) {
+        self.onFastImageLoad(self.onLoadEvent);
+    }
+}
+
+- (void)imageDidLoadObserver:(NSNotification *)notification {
```
evanjmg: This is the line that is crashing for me. @StevenMasini ? any ideas? 
```
Fatal Exception: NSInvalidArgumentException
0  CoreFoundation                 0x208d19eb8 __exceptionPreprocess
1  libobjc.A.dylib                0x207f13f18 objc_exception_throw
2  CoreFoundation                 0x208c2ef10 -[NSOrderedSet initWithSet:copyItems:]
3  CoreFoundation                 0x208d1f924 ___forwarding___
4  CoreFoundation                 0x208d216f0 _CF_forwarding_prep_0
5  App                      0x102ddae0c -[FFFastImageView imageDidLoadObserver:] (FFFastImageView.m:79)
6  Foundation                     0x2097b7618 __57-[NSNotificationCenter addObserver:selector:name:object:]_block_invoke_2
7  CoreFoundation                 0x208c85fcc __CFNOTIFICATIONCENTER_IS_CALLING_OUT_TO_AN_OBSERVER__
8  CoreFoundation                 0x208c85f8c ___CFXRegistrationPost_block_invoke
9  CoreFoundation                 0x208c85434 _CFXRegistrationPost
10 CoreFoundation                 0x208c850c0 ___CFXNotificationPost_block_invoke
11 CoreFoundation                 0x208bfaae4 -[_CFXNotificationRegistrar find:object:observer:enumerator:]
12 CoreFoundation                 0x208c84b44 _CFXNotificationPost
13 Foundation                     0x2096a4b24 -[NSNotificationCenter postNotificationName:object:userInfo:]
14 UIKitCore                      0x236e74ca4 -[UIViewAnimationState sendDelegateAnimationDidStop:finished:]
15 UIKitCore                      0x236e75024 -[UIViewAnimationState animationDidStop:finished:]
16 QuartzCore                     0x20d3d2338 CA::Layer::run_animation_callbacks(void*)
```

evanjmg: perhaps check for source.url?

StevenMasini: > perhaps check for source.url?

Any way to reproduce the crash ? What was the `source.url` ?

StevenMasini: Yeah this crash is link with the observer that I attached to the event.

`[FFFastImageView imageDidLoadObserver:`

I will investigate this. Would be helpful if you could tell me how do you call `<FastImage>`, which props to you give to it.

evanjmg: I still I’m having trouble reproducing as my app has hundreds of images in it and I need to add more metadata to crashalytics. The only  props I use are onLoad, source.headers, source.uri and priority. Perhaps it’s an image resp error?

evanjmg: For now I've reverted my production build to prevent crashes; hopefully I won't get memory leaks with 5.2.0. I'll try to reproduce on staging as I develop new features.

pribeh:
This issue is severe. Can we get this reviewed.

DylanVann:
@StevenMasini This looks good. Turning on some warning checks confirms that you are right about the usage of `self` inside blocks.

I would merge these changes except I'm unsure about the retry handling code. Whatever we provide for retrying should be consistent for Android and iOS.

If you remove the retry related changes I can merge this and we could handle them in another PR.

StevenMasini:
@DylanVann I will extract the auto-retry logic from this PR and submit a new PR specific to it.

I also need to fix the issue found by @evanjmg

ios/FastImage/FFFastImageView.m
StevenMasini: @evanjmg Could you, using XCode intercept the crash and then print the value of `source.url` in the console. I don't know if you are familiar with XCode debugging tools.

[Debugging with XCode](https://developer.apple.com/library/archive/documentation/DeveloperTools/Conceptual/debugging_with_xcode/chapters/debugging_tools.html)

In the breakpoint panel, enable `All Exceptions` to be caught. 
Then once the crash occur, in the debug console type `po` (stand for print object) then the value you want to prospect.

```
po source.url
```

breakpoint-panel

StevenMasini:
@DylanVann I removed the auto-retry logic from this branch.

If @evanjmg come back to me about his crash, I can fix it in another PR.

Also I am willing to contribute more on the native iOS part if you need me 👍

evanjmg:
I had to revert to 5.2.0 due to the images not appearing sometimes (perhaps due to retry logic a can’t seem to figure it out) and also I can’t easily reproduce any crashes but yes they still happen with the old code. For the crash, I’ve added a hacky try catch and check if the method exists before executing and that patches it. Right now, I’ll pull your latest and see if the images not appearing issue goes away.

StevenMasini:
@evanjmg I haven't updated `@steven/react-native-fast-image` as I need the auto-retry logic in my own project.

Where did you put your `try/catch` ?

evanjmg:
`- (void)imageDidLoadObserver:(NSNotification *)notification {
    FFFastImageSource *source = notification.object;
    if (source != nil) {
        BOOL responds = [self respondsToSelector:@selector(sd_setImageWithURL:)];
        if (responds == YES) {
            @try {
                [self sd_setImageWithURL:source.url];
            } @catch (NSException * e) {
                NSLog(@"Exception: %@", e);
            }
        }
    }
}`

evanjmg:
```
- (void)imageDidLoadObserver:(NSNotification *)notification {
    FFFastImageSource *source = notification.object;
    if (source != nil) {
        BOOL responds = [self respondsToSelector:@selector(sd_setImageWithURL:)];
        if (responds == YES) {
            @try {
                [self sd_setImageWithURL:source.url];
            } @catch (NSException * e) {
                NSLog(@"Exception: %@", e);
            }
        }
    }
}
```

StevenMasini:
@DylanVann Yeah I was thinking about it yesterday while removing the auto-retry logic. I will submit a PR directly to SDWebImage for this feature.

But for the notification center, I would still advise you to keep this change as it's mostly RN specific issue. I can't suggest something like this directly to SDWebImage.

The problem with RN is that the component will re-render only if the props have changed. So if a similar image has finish loading in another instance of FastImageView, then previously loaded FastImageView won't be refreshed.

I checked Android don't have this problem for some reason.

PR #441, opened 2019-04-01
Efficient improvement with PureComponent

ifsnow:
As you all know, unnecessary re-rendering can cause performance problems. I think using PureComponent is generally more efficient than not. The more logic is added to the render function, the greater the effect.

Please consider it positively. I hope this helps.


DylanVann:
Making this a pure component may help some people with performance a bit, thanks for the contribution.

In this case I don't want to convert this back into a class component so I'm doing the same thing using `React.memo` - https://github.com/DylanVann/react-native-fast-image/pull/449

Issue #447, opened 2019-04-14
SDWebImage update to 5.x


ivanmoskalev:
Hi! Is there any particular reason to stay on the 4.x version of `SDWebImage` backend? If not, I can perform the update to 5.x. I'd very much like to use `react-native-fast-image` in an app that is being incrementally rewritten using `react-native` – and this app is already using `SDWebImage` 5.0.

dreampiggy:
@ivanmoskalev I create a PR for this. Please help me to contact the maintainer of react native fast image, so that PR can be merged soon.

See: #454

ivanmoskalev:
@DylanVann We need you here.

ivanmoskalev:
@DylanVann Hi! 

If you need help with managing incoming issues, I could help out with that. 24 PRs + 128 issues is a lot, I can understand that. That said, silently disregarding other people's work (like @dreampiggy's #454) seems not the way to go.

If you feel that the #454 is somehow below par, or that #447 is proposing something that you don't like, just give feedback. If you don't have time to review and test, just give feedback. If you like to see it merged, but don't have time, just give feedback. Either way, please give us at least something so we could plan further actions on our own projects that depend on `react-native-fast-image`.

PR #454, opened 2019-04-23
Upgrade SDWebImage 5.0, using SDAnimatedImageView instead of FLAnimatedImageView for better animation support

dreampiggy:
Also, fix the bug of `cacheOnly` behavior on iOS

## Changes
1. The dependency of iOS(SDWebImage) upgrade to 5.0
2. The FLAnimatedImage implementation is dropped, since SDWebImage 5.0 provide built-in solution for animation images, which is better than FLAnimatedImage (Tied to GIF only, no APNG/WebP support), learn more here: [Wiki - Animated Image](https://github.com/SDWebImage/SDWebImage/wiki/Advanced-Usage#animated-image-50)
3. Fix the bug of `cacheOnly` attribute, it should not request any network. However, previously, it does a wrong behavior, which request the network, but only store to memory cache (This option now renamed to `SDWebImageContextStoreCacheType[@(SDImageCacheTypeMemory)]` context option)

## Related
#447


dreampiggy:
@DylanVann Hi there. 
I'm a collaborator of SDWebImage, Could you please have a review for this ?

SDWebImage 5.0 provide many many feature from 4.0, for better customizable, better animation support (A brand new solution for all animated image), better decoder plugins for massive image format. I think it's better to migrate this upstream dependency. 

See more : https://github.com/SDWebImage/SDWebImage/releases/tag/5.0.0

react-native-fast-image.podspec
```diff
@@ -20,8 +20,5 @@ Pod::Spec.new do |s|
   s.exclude_files = "ios/Vendor/**/*.{h,m}"
 
   s.dependency 'React'
-  s.dependency 'SDWebImage/Core'
-  s.dependency 'SDWebImage/GIF'
-  s.dependency 'SDWebImage/WebP'
```
dreampiggy: In 5.0, all external image format (Not supported by Apple's built-in framework), is a coder plugin.

The user, can using another Pod, to register the plugin, don't need a podspec subspec for this. Which make it more modular and under control.

For example, if user want to use WebP format, they can do this:

```
pod 'react-native-fast-image', :path => '../node_modules/react-native-fast-image'
pod 'SDWebImageWebPCoder'
```

Then register the coder plugin, in `AppDelegate.m`

```objectivec
- (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions
{
// ...
// Register WebP format support
[SDImageCodersManager.sharedManager addCoder:SDImageWebPCoder.sharedCoder]
// ...
}
```

DylanVann: This would mean that by default this module would no longer support as many formats though right?

dreampiggy: > This would mean that by default this module would no longer support as many formats though right?

Whether SDWebImage supports some format, or not. It's now, totally controlled by the users. If users want, they can supports all the format, listed in : [Coder Plugin List](https://github.com/SDWebImage/SDWebImage/wiki/Coder-Plugin-List).

The main repo (or `SDWebImage` pod itself), only supports the built-in format by Apple's SDK. Typically the JPEG/PNG/TIFF/GIF/HEIF, etc.

And I see, react-native-fast-image SDK, didn't rely on any dependency on what original image format is. Isn't it ? We just care about the animation, and their meta info (loop count/ frame count), but not what the original compressed data is.

dreampiggy:
Hi. Seems this PR is update to the master branch. And I've answered that question. Any new update about this ?
@DylanVann Do you have time recently ? If there are anything I can do to make this merged, please feel free to reply.

DylanVann:
I think this is good assuming it supports formats we used to support without manual steps. Just need time to test it out, maybe tonight. Thank you!

DylanVann:
Had to make some modifications but it's in. I wish there was a better process for this kind of thing using semantic release...

Thank you @dreampiggy !

PR #455, opened 2019-04-23
Fix:failed to load image while RN rebuilding view hierachy

bryanlau330:
While RN rebuilding view hierachy,the LayoutParams of ImageViews have been reset.Glide refuse to do further process if overrideWidth=0 or overrideHeight=0.

### DIFF
--- a/android/src/main/java/com/dylanvann/fastimage/FastImageViewManager.java
+++ b/android/src/main/java/com/dylanvann/fastimage/FastImageViewManager.java
@@ -102,7 +102,7 @@ class FastImageViewManager extends SimpleViewManager<FastImageViewWithUrl> imple
                     //    - android.resource://
                     //    - data:image/png;base64
                     .load(imageSource.getSourceForLoad())
-                    .apply(FastImageViewConverter.getOptions(source))
+                    .apply(FastImageViewConverter.getOptions(source).override(view.getWidth(), view.getHeight()))
                     .listener(new FastImageRequestListener(key))
                     .into(view);
         }


android/src/main/java/com/dylanvann/fastimage/FastImageViewManager.java
```diff
@@ -102,7 +102,7 @@ public void setSrc(FastImageViewWithUrl view, @Nullable ReadableMap source) {
                     //    - android.resource://
                     //    - data:image/png;base64
                     .load(imageSource.getSourceForLoad())
-                    .apply(FastImageViewConverter.getOptions(source))
+                    .apply(FastImageViewConverter.getOptions(source).override(view.getWidth(), view.getHeight()))
```
DylanVann: Could you add a comment explaining this? Also how would I go about reproducing this issue?

bryanlau330: I am working hard on my English,I'm sorry if you can not get the point of what I said below.In one of my released APPs, I am using a FlatList to display a web photo album. The images in the album are downloaded and displayed using FastImage.FastImage works perfectly when I first enter the album page, but if I reorder the album pictures and reload FlatList, FastImage will fail to load.I debugged, and found that the failure was in the Glide library. The source code of Glide showed that when the width or height of the LayoutParams of a view is 0, it returned the failure directly.Reloading FlatList causes the setSrc method of FastImageViewManager called, but at that time, the LayoutParam of ImageViews is not meeting the Glide's requirements.

PR #459, opened 2019-04-26
delete from cache by URL

ARDcode:
This code add to use **FastImage.deleteImage(url)** method on ios and android.
It useful, when you want update image with same url or for more cache control.

Example:
```
FastImage.deleteImage(url).then(deletedImagePath => {
                    console.log(deletedImagePath);
};
```
Its async method.

### DIFF
--- a/android/src/main/java/com/dylanvann/fastimage/FastImageViewModule.java
+++ b/android/src/main/java/com/dylanvann/fastimage/FastImageViewModule.java
@@ -1,15 +1,23 @@
 package com.dylanvann.fastimage;
 
 import android.app.Activity;
+import android.support.annotation.Nullable;
 
 import com.bumptech.glide.Glide;
+import com.bumptech.glide.load.DataSource;
+import com.bumptech.glide.load.engine.GlideException;
 import com.bumptech.glide.load.model.GlideUrl;
+import com.bumptech.glide.request.RequestListener;
+import com.bumptech.glide.request.target.Target;
+import com.facebook.react.bridge.Promise;
 import com.facebook.react.bridge.ReactApplicationContext;
 import com.facebook.react.bridge.ReactContextBaseJavaModule;
 import com.facebook.react.bridge.ReactMethod;
 import com.facebook.react.bridge.ReadableArray;
 import com.facebook.react.bridge.ReadableMap;
 import com.facebook.react.views.imagehelper.ImageSource;
+
+import java.io.File;
 
 class FastImageViewModule extends ReactContextBaseJavaModule {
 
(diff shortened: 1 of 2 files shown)


android/src/main/java/com/dylanvann/fastimage/FastImageViewModule.java
```diff
@@ -45,12 +53,39 @@ public void run() {
...
+        Glide
+                .with(activity.getApplicationContext())
+                .asFile()
+                .load(url)
+                .listener(new RequestListener<File>() {
+                    @Override
+                    public boolean onLoadFailed(@Nullable GlideException e, Object model, Target<File> target, boolean isFirstResource) {
+
+                        return false;
+                    }
```
Ilphrin: I don't know much of Java, but shouldn't you reject the Promise if the `onLoadFailed` is called?

Ilphrin (approved):

android/src/main/java/com/dylanvann/fastimage/FastImageViewModule.java
ARDcode: Oh, sorry, u right. I will update request, thx

ios/FastImage/FFFastImageViewManager.m
```diff
@@ -33,5 +34,13 @@ - (FFFastImageView*)view {
     [[SDWebImagePrefetcher sharedImagePrefetcher] prefetchURLs:urls];
 }
 
+RCT_EXPORT_METHOD(deleteImage:(NSString *)url)
+{
+    NSURL *nsurl = [NSURL URLWithString:url];
+    SDWebImageManager *imageManager = [SDWebImageManager sharedManager];
+    NSString *cacheKey = [imageManager cacheKeyForURL:nsurl];
+    [[SDImageCache sharedImageCache] removeImageForKey:cacheKey fromDisk:TRUE withCompletion:0];
+}
```
dreampiggy: Please note that SDWebImage's image cache have Async version API to delete the cache. And the JavaScript API also using Promise. So the better and robust code of this can be:

```objective-c
RCT_EXPORT_METHOD(deleteImage:(NSString *)url
                     resolver:(RCTPromiseResolveBlock)resolve
                     rejecter:(RCTPromiseRejectBlock)reject)
{
    NSURL *nsurl = [NSURL URLWithString:url];
    SDWebImageManager *imageManager = [SDWebImageManager sharedManager];
    NSString *cacheKey = [imageManager cacheKeyForURL:nsurl];
    NSString *cachePath = [[SDImageCache sharedImageCache] cachePathForKey:cacheKey];
    [[SDImageCache sharedImageCache] removeImageForKey:cacheKey fromDisk:YES withCompletion:^{
        resolve(cachePath);
    }];
}
```

Note: SDWebImage on iOS, does not populate the Error during deleting cache, since it's trivial for most users. But if you really want the error info, you need to use the `NSFileManager` and delegate for SDImageCache. But it's little hard to do so (You need to keep a alive instance and pass the context from React to keep the error info)

dreampiggy: Acutally, does the promise fulfilled value `cachePath` means anything ? What if the cache is in memory only but not disk (the `cachePath` will be nil, but promise is still fulfilled)

terencehh:
is there an update on this? this is something that will be very useful for me. I need a way to delete/refresh the cache for a single image url instead of being forced to delete everything via `FastImage.clearDiskCache` and `FastImage.clearMemoryCache`

PR #463, opened 2019-04-30
UPDATE - proguard rules on readme

FelipePaiva1992:
When proguard is enable, the app return this error:
`java.lang.IllegalArgumentException: Unable to find GlideModule implementation`

It`s a common issues, so I just add the fix on Readme file:
```
-keep public class * implements com.bumptech.glide.module.GlideModule
-keep public class * extends com.bumptech.glide.module.AppGlideModule
-keep public enum com.bumptech.glide.load.ImageHeaderParser$** {
  **[] $VALUES;
  public *;
```


DylanVann:
Nice, thank you!

PR #478, opened 2019-05-20
fix proguard rules

pdavid0:
add missing `}`.

Fixes

```
java.io.IOException: proguard.ParseException: Expecting class member description or closing '}' before end of line
```


DylanVann:
Thank you!

PR #479, opened 2019-05-22
Add cache to typings

RWOverdijk:


src/index.d.ts
```diff
@@ -82,6 +82,7 @@ export interface OnProgressEvent {
 
 export interface FastImageProperties {
     source: FastImageSource | number
+    cache?: Cache
```
DylanVann: Cache is not a property of `FastImage`, it's a property of a `FastImageSource`: https://github.com/DylanVann/react-native-fast-image#sourcecache-enum

RWOverdijk: Hah... I think I missed the dot because the readme uses the same heading for all props. Silly me, thanks.

DylanVann: No problem 😄

PR #487, opened 2019-05-31
Add tintColor to typings

mamoru0217:
added tintColor definition 😄


DylanVann:
Thanks @mamoru0217!

Probably going to convert this repo to TypeScript soon and end these issues for good...

Issue #476, opened 2019-05-15
WebP with 6.0.0


pontusab:
- React Native version: `v0.59.8`
- React version: `16.8.6`
- React Native Fast Image version: `6.0.0`

In the latest release using cocapods and WebP the images are not displayed at all. Downgrading to `5.4.2` works just fine. 

I guess there is something with `SDWebImage` that needs to be patched?

guhungry:
Please refer to https://github.com/DylanVann/react-native-fast-image/pull/454#discussion_r277512690

pontusab:
Yes, tried that. Did not work, did it work for you then @guhungry ?

guhungry:
Nope, I didn't use webp in my app.

Ilphrin:
Seems like `SDWebImage` have its support disabled by default (See https://stackoverflow.com/questions/18012527/how-to-load-a-webp-image-with-sdwebimage ) Maybe you can try something like one of the answers @pontusab ?

dreampiggy:
@pontusab See [SDWebImageWebPCoder](https://github.com/SDWebImage/SDWebImageWebPCoder) Readme.

For 5.0, all the external image format coders, are standalone modules (Pods). If you want to use WebP, you can add the Pods and register it in `AppDelegate.m`.


```
pod 'react-native-fast-image'
pod 'SDWebImageWebPCoder'
```

```
- (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions
{
    // ...
    // Register WebP format support
    [SDImageCodersManager.sharedManager addCoder:SDImageWebPCoder.sharedCoder];
    // ...
}
```

pontusab:
Thanks, I got it to work 🥇

pontusab:
Import in `AppDelegate.m`:
```
#import "SDImageCodersManager.h"
#import <SDWebImageWebPCoder/SDImageWebPCoder.h>

```

An add:

```
- (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions
{
    // ...
    // Register WebP format support
    [SDImageCodersManager.sharedManager addCoder:SDImageWebPCoder.sharedCoder];
    // ...
}
```

aibrahim3546:
@pontusab What is the step you do to get it works?

I did like this

- add `pod 'SDWebImageWebPCoder'`
- run `pod install`

and then in `AppDelegate.m`

```
[...]

PR #488, opened 2019-06-02
Add the Demo for iOS about how to use WebP format for 6.0.0 version

dreampiggy:
Seems after I create the PR for upgrating the iOS integrate dependency [SDWebImage](https://github.com/SDWebImage/SDWebImage) to 5.0.0 via #454, some people find it hard to start how to use WebP format in v6.0.0 version. Like #476

So, I update the current demo. To show how to integrate your exist React Native project, with the 6.0.0 version, to support external image format used by SDWebImage. SDWebImage now support external format beyond the limit of Apple's system library. Check all of available external image format [here](https://github.com/SDWebImage/SDWebImage/wiki/Coder-Plugin-List)

This PR change only iOS demo, contains both two cases:

+ `react-native-fast-image-example`, don't use CocoaPods. User who need WebP on iOS, should consider to use [Carthage](https://github.com/Carthage/Carthage), or manual clone the Git submodule and add the library. This time I use the Git submodule, but it's recommanded to use Carthage instead (another CLI util on iOS development), which is easy to use if you're not familiar to Xcode build settings.
+ `react-native-fast-image-example-cocoapods`, use CocoaPods. The most easy way to use WebP. Just add one line `pod 'SDWebImageWebPCoder'`

I run both of two examples locally and it works.

@DylanVann Maybe we need some wiki or readme, to show how to add new image format support ? Does Android's Glide support external format like iOS's SDWebImage ?

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,9 @@
 [submodule "ios/Vendor/SDWebImage"]
 	path = ios/Vendor/SDWebImage
 	url = https://github.com/rs/SDWebImage.git
+[submodule "react-native-fast-image-example/ios/Vendor/SDWebImageWebPCoder"]
+	path = react-native-fast-image-example/ios/Vendor/SDWebImageWebPCoder
+	url = https://github.com/SDWebImage/SDWebImageWebPCoder.git
+[submodule "react-native-fast-image-example/ios/Vendor/libwebp-Xcode"]
+	path = react-native-fast-image-example/ios/Vendor/libwebp-Xcode
+	url = https://github.com/SDWebImage/libwebp-Xcode.git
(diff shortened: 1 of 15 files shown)


DylanVann:
That's weird. I tested WebP on that PR and it worked without modifications. Maybe a caching issue 😆

dreampiggy:
@DylanVann That PR did not update the Demo. It still using the react-native-fast-image 5.x version.
WebP is not supported by Apple's built-in framework. It must incldue a external third party library, which increase the binary size. This is why we remove that dependency in SDWebImage 5.x version. Make it modular.

DylanVann:
I understand. Glide can display WebP if the system can, similar situation to SDWebImage.

I think for now people would be better off sticking to well supported image formats.

PR #491, opened 2019-06-05
Round the corners manually on Nougat [Android]

lukefanning:
Fix has been done in other forks. Just performed the same fix and made it Nougat specific with the latest master

Edit: Improved the fix by allowing override of width and height on the resulting image so that the border radius ratio is correct. This does unfortunately have the effect of potentially losing image quality since you lose the benefits of downscaling


DylanVann:
I can't see issues with current corners on recent android/lib versions:

Screenshot_1631735073

DylanVann:
I'm going to close https://github.com/DylanVann/react-native-fast-image/pull/491 and https://github.com/DylanVann/react-native-fast-image/pull/411 . If it's still an issue please reopen, thank you!

PR #496, opened 2019-06-19
Use device scale when loading images

tberman:
This fixes retina issues I was having locally.


DylanVann:
Awesome, thank you!

DylanVann:
@tberman I have to revert this because it prevents gifs from working. Could you also give a more detailed description of the issue you're experiencing?

tberman:
So, the issue is that the image gets downloaded and then displayed on a canvas that is far too small for the image, cropping it always.

For the gifs, there are two fixes, the easy one is to retain the old behavior if its a gif or other animated image, but the real fix is going to be resizing it frame by frame.

https://github.com/SDWebImage/SDWebImage/issues/1210 and other searches for 'SDImage retina' show the issue and make it clear the fix should be userland (which confuses me)

PR #511, opened 2019-07-04
Fixed IllegalArgumentException crash when context was instance of Con…

leanmazzu:
Hi! This is a fix that we at Skyscanner needed to implement in order to avoid an `IllegalArgumentException: You cannot start a load for a destroyed activity` crash.

The solution is to add an extra check to consider the case when the `baseContext` is an instance of `ContextWrapper`. 

As it was before it was never calling the `isActivityDestroyed()` method as neither of the checks of `if (instance of...` was evaluating to true.


DylanVann:
Looks awesome, thank you @leanmazzu !

PR #526, opened 2019-07-24
Update Readme for RN 0.59.x or lower

yogeshkotadiya:
Added Installation steps for adding RN-fast-image on RN version 0.59.x or lower

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,14 +55,34 @@ and
 -   [x] GIF support.
 -   [x] Border radius.
 
-## Usage
+## Getting Started
 
-**Note: You must be using React Native 0.60.0 or higher to use the most recent version of `react-native-fast-image`.**
+**1. Installation**
+
+**For React Native 0.60.0 or higher.**
 
 ```bash
+npm install react-native-fast-image
+
+# OR
+
 yarn add react-native-fast-image
 ```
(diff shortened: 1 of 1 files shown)


DylanVann:
I don't think the benefits of backwards compatibility / documenting backwards compatibility are worth the cost given the pace of React Native development changing. Sorry if it's inconvenient.

WeslleyNasRocha:
I find this pretty valid, i use this on my company app, and it cant keep the pace of the react-native updates, we currently updated from 0.54 to the 0.59 and not planning update to the 0.60 until its matured.

WeslleyNasRocha:
Actually, a lot of libraries do this.

Kris-Lee-SH:
So, version 6.1.1 is for RN 0.59?

cupcakearmy:
As a single maintainer I totally understand that the cost of making it backwards compatible are not feasible. BUT just a hint to wich version was "last seen working with 0.59 .e.g" would be very useful.
As a company you don't always have the luxury of upgrading to every latest version when you have tons of native dependencies.

Seems like 6.1.1 for 0.59

Thanks for the library though! ❤️

xclidongbo:
ThankX

PR #541, opened 2019-08-13
Glide to fresco adapter

OleksandrKucherenko:
Create a special library that allows switching of the solution from Glide to Fresco lib.


OleksandrKucherenko:
@DylanVann ^ can you take a look? will be happy to merge those changes and make them available for all devs.

DylanVann:
React Native's default image component uses Fresco. What's the advantage of using this?

OleksandrKucherenko:
> React Native's default image component uses Fresco. What's the advantage of using this?

This package/module/library uses Glide... I just provide an alternative that allows to drop glide dependency and redirect all calls to Fresco

DylanVann:
Since this project was created to avoid using Fresco I don't think this is the right route to go.

I'd really like to see some project that benchmarks Android image loading libraries and provides a feature matrix. Then we could make a well informed decision on what is the best library, and use that.

OleksandrKucherenko:
https://medium.com/hackernoon/picasso-universal-image-loader-or-glide-that-s-the-question-af34fa7f5e63

DylanVann:
I have looked for benchmarks. I’m not particularly interested in if any
library uses slightly more memory, or slightly more methods, I’m more
concerned about speed and reliability in terms of loading from cache, which
is more difficult to measure.

Choosing Glide was a somewhat subjective decision. It seems to me that it
more reliably loads from cache and loads quickly. It would be preferable to
have numbers on this though.

OleksandrKucherenko:
Anyway, I just give an option for those who don't want to mix several different libraries in project. Fresco is quite common for RN projects. 

option give ability to drop dependency on glide and reuse the Fresco lib. By functionality, they are almost the same.

PR #542, opened 2019-08-13
Added Support For Animated Webp

theFong:
Uses [this integration library](https://github.com/zjupure/GlideWebpDecoder) to support animated webps.


android/build.gradle
```diff
@@ -58,6 +58,8 @@ dependencies {
     implementation "androidx.core:core:1.0.0"
     implementation "androidx.appcompat:appcompat:1.0.0"
     implementation "androidx.annotation:annotation:1.0.0"
+    // webpdecoder
+    implementation "com.zlc.glide:webpdecoder:1.5.${_glideVersion}"
```
OleksandrKucherenko: I just think maybe it's better to update documentation and show how to attach support of extra image formats... otherwise, it's just the increase of dependencies list in any project, even if you do not use the webp format at all.

bigorangemachine:
This would be a great feature to have within the library as it'll have feature parity with iOS!

I don't think the bloat is an issue as most react native projects are pretty bloated as is

PR #544, opened 2019-08-21
add caching management

bpouzet:
retry guhundry pull request  #425 with latest version of fastimage

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -145,6 +145,10 @@ interface FastImageStatic extends React.ComponentClass<FastImageProperties> {
     }
 
     preload(sources: FastImageSource[]): void
+	
+	clearMemoryCache(): void
+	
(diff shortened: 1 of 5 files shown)


mrousavy:
@DylanVann could you please take a look at this PR? Clearing cache can be quite beneficial 😬

jscarmona (approved):

src/index.d.ts
```diff
@@ -124,6 +125,14 @@ export interface FastImageProperties {
     testID?: string
 }
 
+export interface PreloadProgressHandler {
+    (loaded: number, total: number): void
+}
+
+export interface PreloadCompletionHandler {
+    (loaded: number, skipped: number): void
```
DylanVann, line 133: Ideally we should be able to tell the status of individual images with these callbacks. I think SDWebImage might support this now.

Issue #534, opened 2019-08-05
cocoapod install error


Ekhui:
when I run pod 'RNFastImage', :podspec => '../node_modules/react-native-fast-image/RNFastImage.podspec'

it will be :

[!] Unable to find a specification for `SDWebImageWebPCoder (~> 0.2.3)` depended upon by `RNFastImage`

AbdulmalickDimnang:
@Ekhui just go to ios and then "pod repo update" after that, try again "pod install" then it should be working fine

drminh2807:
> @Ekhui just go to ios and then "pod repo update" after that, try again "pod install" then it should be working fine

It worked!

Nathan-Patnam:
For me I had the exact same file paths being used for my dependencies in both the 'Copy Bundle Resources" and the 'Copy Pod Resources' So I deleted everything in my Copy Pod Resources to have something like:


![Screen Shot 2019-10-13 at 2 09 10 AM](https://user-images.githubusercontent.com/17086697/66711692-b1f76500-ed5e-11e9-8a22-02ffc8f1a26b.png)
![Screen Shot 2019-10-13 at 2 09 15 AM](https://user-images.githubusercontent.com/17086697/66711694-b459bf00-ed5e-11e9-98c8-99dfebd39151.png)

Issue #257, opened 2018-08-01
IOS invariant violation: requireNativeComponent FastImageView was not found in the UI manager


suupham2506:
how can I fix this error??????

SurpassRabbit:
I have the same problem, I have the correct `link`

predescu:
I had the same issue after I install and link the library.
I realised that I forgot to rebuild the project. Maybe is the same case for you.

`react-native run-ios`

jan-happy:
I also have the same problem. Rebuilt, tried manual linking and Cocoapods. Always with the same result: invariant violation: 

> Error: Invariant Violation: requireNativeComponent: "FastImageView" was not found in the UIManager.

antonijap:
I have this problem as well.

objectiveSee:
Make sure you run `pod install` and re-build your entire project.

a-eid:
I'm facing the same issue, building the project with xcode yield the following error: 
![Screen Shot 2019-04-11 at 2 14 53 AM](https://user-images.githubusercontent.com/19273413/55922072-e481a300-5bff-11e9-9651-52656154af81.png)

steps I did to install the lib. 
- `npm install `
- `react-native link`
- `cd ios; pod install`
- add xcode project to lib and add `libFastImage.a` to build phases.

nikoremi97:
Stille not working for me @a-eid . I cant find libFastImage.a Do you have any ideas?

zdQAQ:
I have tried `react-native link react-native-fast-image` which set `pod 'react-native-fast-image', :path => '../node_modules/react-native-fast-image'` in my Podfile and it doesn't work for some reason.
then **i follow the manual link way which works** as [https://github.com/DylanVann/react-native-fast-image/blob/master/docs/installation-manual.md#ios](url) shows.

In Xcode:

- Add FastImage.xcodeproj to Libraries.
- Add libFastImage.a to Link Binary With Libraries under Build Phases.

amit2908:
I am working on Xcode 10.2.1 and while adding FastImage.xcodeproj to Libraries I am getting a Xcode Crash every time.

xxRockOnxx:
Why was this commited? https://github.com/DylanVann/react-native-fast-image/commit/41de1fdc6d241d20ce77c5422b9a72d0e433e16a

yurtaev:
[...]

PR #553, opened 2019-09-05
Documenting quick fix pointed out in issues #534 and in #257

espipj:
It helps when you migrate to the newer versions of react-native-fast-image which support 60.0 from older versions
https://github.com/DylanVann/react-native-fast-image/issues/534#issuecomment-519015141


DylanVann:
Awesome, thank you @espipj!

PR #563, opened 2019-09-18
Add react-native link command to README

evancloutier:
Looks like this instruction was missing from README -- added back in.

Feel free to add further clarification for lower versions of RN (assuming that the statement about latest working on 0.60+ is adequate).

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,10 @@ and
 yarn add react-native-fast-image
 ```
 
+```bash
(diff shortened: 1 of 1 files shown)


DylanVann:
The link command is not needed in the current React Native version. Autolinking is used now.

https://facebook.github.io/react-native/blog/2019/07/03/version-60

PR #570, opened 2019-10-01
Performance improvements in SDWebImage

gtfunes:
- Set maxBufferSize to 1MB to improve GIF loading
- Add SDWebImageScaleDownLargeImages by default

### DIFF
--- a/ios/FastImage/FFFastImageView.m
+++ b/ios/FastImage/FFFastImageView.m
@@ -148,7 +148,7 @@
         }];
         
         // Set priority.
-        SDWebImageOptions options = SDWebImageRetryFailed || SDWebImageScaleDownLargeImages;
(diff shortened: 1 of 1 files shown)


ngochiencse:
I think this change is neccessary. Why did you close?

gtfunes:
hi @ngochiencse, I did these specific changes for a project I was working on that is now deprecated. I agree, a similar change is necessary (not sure if something like this was added since 2019, haven't checked the repo in a while) but it would need to be configurable, I would also check is something similar could be done for Android, etc. I might work on this again in the future, but not for now. If you need these specific changes in your project you can always apply them using [patch-package](https://www.npmjs.com/package/patch-package) (which I didn't know about back then).

PR #586, opened 2019-10-16
Allow iOS to clear the tintColor property.

sleeper-seth:
By allowing the image color to be set to nil, it's possible to set and then unset a tint color.

This now matches the Android behavior defined in FastImageViewManager.java

[FastImageViewManager](https://github.com/DylanVann/react-native-fast-image/blob/master/android/src/main/java/com/dylanvann/fastimage/FastImageViewManager.java#L116)

### DIFF
--- a/ios/FastImage/FFFastImageView.m
+++ b/ios/FastImage/FFFastImageView.m
@@ -61,16 +61,16 @@
 }
 
 - (void)setImageColor:(UIColor *)imageColor {
-    if (imageColor != nil) {
-        _imageColor = imageColor;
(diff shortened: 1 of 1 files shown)


eminsr:
This doesn't really works. It removes the `tintColor` but image still has a black `tintColor` for some reason.

Noitham:
There are still issues with the`tintColor` on iOS devices, the value is not cleared, so if switching on/off, the value stacks up.

DylanVann:
Thank you! Fixed in 8.6.14 (#1098). I kept the untinted image so the tint can be removed, since the tinted copy replaces the original.

Issue #470, opened 2019-05-09
Fallback prop with cache


peckpeck20:
i've replaced all Image components with FastImage in a relative mid-large app. I've come to notice that in some occasions I will get a black or blank image throughout the app. Reading the docs I decided to add the Fallback prop.

Adding the Fallback prop to the FastImage component causes an error with the cache control.
This is how my component looks at the moment, if I remove source cache the error goes away, does this mean that if I use the fallback option to the native Image component I can't set a cache pref. 

Could someone elaborate the proper use the component with the fallback prop. I was thinking I could maybe create a wrapper component to accomplish this, triggering the fallback prop only when the FastImage has an onError.  any suggestions would be much appreciated  :)

`                <FastImage
                  source={{
                    uri: productImage,
                    priority: FastImage.priority.high,
                  // cache: FastImage.cacheControl.web,
                  }}
                  style={itemStyles.image}
                  resizeMode={FastImage.resizeMode.contain}
                  fallback
                />
`

**Expected behavior**
If the FastImage component has errors or doesn't load then fall back to the native Image component

**Screenshots**

<img width="475" alt="Screenshot 2019-05-09 at 13 38 54" src="https://user-images.githubusercontent.com/22203676/57447540-02bedb00-7260-11e9-8396-2491f46f05fc.png">

<img width="475" alt="Screenshot 2019-05-09 at 13 41 16" src="https://user-images.githubusercontent.com/22203676/57447599-22ee9a00-7260-11e9-95c8-156c85e0eb0d.png">

**Dependency versions**

- React Native version:  ^0.58.5
- React version: ^16.6.3
- React Native Fast Image version: ^ 6.0.0

DylanVann:
Fixed in https://github.com/DylanVann/react-native-fast-image/commit/ba0f238821ba23517b8e62e759f685b8cd67c0c6 . Thank you!

PR #590, opened 2019-10-22
fix: remove cache property if using fallback

milansusnjar:
iOS image has its own properties for cache and if fallback is true and regular image is being rendered we can't set FastImage's cache value because it doesn't match.

It closes https://github.com/DylanVann/react-native-fast-image/issues/470

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -23,16 +23,17 @@ function FastImageBase({
     forwardedRef,
     ...props
 }) {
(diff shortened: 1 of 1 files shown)


DylanVann:
@milansusnjar Thanks for this! This got merged in this commit https://github.com/DylanVann/react-native-fast-image/commit/ba0f238821ba23517b8e62e759f685b8cd67c0c6 .

PR #594, opened 2019-11-06
Fix and update Gradle setup

friederbluemle:
Three commits here:

The first one changes the Gradle script to include the Android Gradle plugin _conditionally_. That way, it's only used when the `android` folder is opened individually, and it does not interfere with consumer projects using this library. It also updates the plugin to `3.5.3`.

See [this thread](https://github.com/facebook/react-native/pull/25569), and [this comment](https://github.com/facebook/react-native/pull/25569#issuecomment-532504277) for details and more information.

The second commit updates `glide` from `4.9.0` to `4.10.0`. The newer version drops a dependency on the deprecated support libs, avoiding a Lint error.

Finally the third commit fixes the only remaining Lint issue in the Android code, by replacing `ImageView` with `AppCompatImageView` from the AndroidX libraries.

With these changes, the Android code now builds without any Lint issues! :+1:

<sub><sup>Once approved, please use a *normal* GitHub merge (i.e. **NO rebase/squash** merge) to integrate the commit(s) from the PR head branch. The changes are broken up into meaningful, [atomic commits](https://www.freshconsulting.com/atomic-commits/), and my branch should already be up-to-date with the latest base branch. If it isn't, or if you want me to change anything, please let me know, and I will update the branch as soon as possible. Thank you!</sup></sub>

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -57,7 +57,7 @@ repositories {
     jcenter()
 }
 
-def glideVersion = safeExtGet('glideVersion', '4.9.0')
+def glideVersion = safeExtGet('glideVersion', '4.10.0')
 
 dependencies {
     //noinspection GradleDynamicVersion
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -1,47 +1,42 @@
-buildscript {
-    repositories {
-        jcenter()
-        google()
-    }
-    dependencies {
-        classpath "com.android.tools.build:gradle:3.4.0"
-    }
-}
-
-apply plugin: "com.android.library"
-
 def safeExtGet(prop, fallback) {
     rootProject.ext.has(prop) ? rootProject.ext.get(prop) : fallback
 }
 
-def _excludeAppGlideModule = safeExtGet("excludeAppGlideModule", false)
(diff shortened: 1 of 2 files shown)


SaeedZhiany:
@friederbluemle 

I suggest putting a reference to [this thread](https://github.com/facebook/react-native/pull/25569) and especially [your own comment.](https://github.com/facebook/react-native/pull/25569#issuecomment-532504277) when applying the change. by that, we can keep a list of repositories that the change (conditionally loading of AGP) has been applied to them, so in case of a problem that related to this change, we can easily find the repositories and fix them. also, it can be used as proof that the change is correct and the conversation on that thread can be useful for anyone that doesn't know why exactly this change is needed.

> It also updates the plugin to 3.5.2 since Android Studio 3.5.2 is now available in the Stable channel.

Please edit this one, actually, it's upgraded to 3.5.3 in your PR

friederbluemle:
Yes, good idea. Thanks @SaeedZhiany 👍

DylanVann:
Resolved conflicts and merged into main.

friederbluemle:
Thanks @DylanVann :+1:

PR #601, opened 2019-11-16
CocoaPods - Add Tag Version

ptlu:
I faced an issue today where I attempted to install a previous specific version of `react-native-fast-image` into my codebase.  I saw that the version installed by CocoaPods did not match the version in my `node_modules` directory – it was pulling the HEAD version always.  This resolves that issue.

Why?
* This change will make the CocoaPods version which is installed match the NodeJS version used.  
  Currently, CocoaPods will install the latest version on head which could not match the JavaScript
  version of the library.

Changes:
1. Use the version tag to match the version installed by CocoaPods to that coming the from the `package.json` file.


ptlu:
Hello! Any update here?

ptlu:
Great thanks!

PR #614, opened 2020-01-04
Load Android Gradle Plugin conditionally

SaeedZhiany:
This wraps the Android Gradle plugin dependency in the buildscripts section of android/build.gradle in a conditional:

```
if (project == rootProject) {
    // ... (dependency here)
}
```

The Android Gradle plugin is only required when opening the project stand-alone, not when it is included as a dependency. By doing this, the project opens correctly in Android Studio, and it can also be consumed as a native module dependency from an application project without affecting the app project (avoiding unnecessary downloads/conflicts/etc).

for more info, you can refer to [this thread](https://github.com/facebook/react-native/pull/25569) and especially [this comment.](https://github.com/facebook/react-native/pull/25569#issuecomment-532504277)

Also, according to [this answer on stackoverflow](https://stackoverflow.com/a/46467800/6340959) `google()` and `maven { url 'https://maven.google.com' }` are pointing to the same repository and no need to have both of them together. so I removed the second one.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -1,10 +1,16 @@
 buildscript {
-    repositories {
-        jcenter()
-        google()
-    }
-    dependencies {
-        classpath "com.android.tools.build:gradle:3.4.0"
+    // The Android Gradle plugin is only required when opening the android folder stand-alone.
+    // This avoids unnecessary downloads and potential conflicts when the library is included as a
+    // module dependency in an application project.
+    if (project == rootProject) {
+        repositories {
+            google()
+            jcenter()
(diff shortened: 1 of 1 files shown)


friederbluemle:
Thanks @SaeedZhiany - This (as well as a couple of other things) is already taken care of in #594

SaeedZhiany:
Hi @friederbluemle, I'm closing this in favor of your PR, just please take a look at [my comment](https://github.com/DylanVann/react-native-fast-image/pull/594#issuecomment-574498694) on your PR.

PR #619, opened 2020-01-08
Use React Native shared CookieJar on Android

charpeni:
On Android, we could reuse the same CookieJar that React Native uses, so we won't have to append authorization headers manually.

The usage is pretty simple, we just have to make sure we got a `Set-Cookie` response in React Native, then those cookies will be automatically appended to matching requests with `react-native-fast-image`.

### DIFF
--- a/android/src/main/java/com/dylanvann/fastimage/FastImageCookieJar.java
+++ b/android/src/main/java/com/dylanvann/fastimage/FastImageCookieJar.java
@@ -1,0 +1,38 @@
+package com.dylanvann.fastimage;
+
+import android.webkit.CookieManager;
+
+import java.util.Collections;
+import java.util.List;
+
+import okhttp3.Cookie;
+import okhttp3.CookieJar;
+import okhttp3.HttpUrl;
+
+public class FastImageCookieJar implements CookieJar {
+    private CookieManager cookieManager;
+
+    private CookieManager getCookieManager() {
+        if (cookieManager == null) {
+            cookieManager = CookieManager.getInstance();
+        }
+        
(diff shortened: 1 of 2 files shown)


android/src/main/java/com/dylanvann/fastimage/FastImageCookieJar.java
```diff
@@ -0,0 +1,38 @@
...
+        if (cookieManager == null) {
+            cookieManager = CookieManager.getInstance();
+        }
+        
+        return cookieManager;
+    }
+
+    @Override
+    public void saveFromResponse(HttpUrl url, List<Cookie> cookies) {
+        // Do nothing
```
DylanVann, line 25: Should we not save new cookies?

charpeni: I can't remember the context of this pull request, but I would say that I don't expect `FastImageCookieJar` to be setting new cookies from a response. If we need an authentication prior to image requests, it should be done from React Native, and then we'll be reusing React Native's CookieJar in FastImage's OkHttpClient.

DylanVann: I know some cookie based auth systems will occasionally refresh a cookie using a response. I'm wondering if any of them would ever refresh a cookie on an image response. You're right it's probably unlikely.

DylanVann: I'm guessing SDWebImage already works like this, do you know if that's correct? Would be good to work the same.

PR #625, opened 2020-01-24
Example whith styled-components

edisonAlbaSarmiento:
An example of how to use fast-image with styled components was added  was README


DylanVann:
Why do you think this needs to be documented within this repo? It doesn't seem that it's non-obvious.

DylanVann:
This could be documented somewhere else, I think it's not related enough to `react-native-fast-image`.

PR #630, opened 2020-02-11
Allow FastImageView component to be the target of the forwardRef

grean:
So, FastImage component can be join by ref in users code.

### DIFF
--- a/RNFastImage.podspec
+++ b/RNFastImage.podspec
@@ -16,6 +16,6 @@ Pod::Spec.new do |s|
   s.source_files  = "ios/**/*.{h,m}"
 
   s.dependency 'React'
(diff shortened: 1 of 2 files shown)


todorone:
LGTM
@DylanVann Can we merge this?

DylanVann:
What does this allow that the current ref does not? I would think having the outermost view would be most useful.

DylanVann:
If you still think this is a good change please open another PR with a description of why the ref change is useful. Thank you!

maksimgorodilov:
Hello @DylanVann this PR can help us to programatically change source uri of the image (for example updating it on scroll) with `setNativeProps`

PR #654, opened 2020-03-22
Fix Source's Cache type

kulyk:
`Cache` property type was equal to `Priority` (with `low | normal | high` values)
However, `FastImage.cacheControl` contains `immutable`, `web` and `cacheOnly` values, so the types are not matching

Fixed by changing `Cache` type to `'immutable' | 'web' | 'cacheOnly'`

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -31,7 +31,7 @@ const priority = {
     high: 'high',
 } as const
 
-type Cache = 'low' | 'normal' | 'high'
+type Cache = 'immutable' | 'web' | 'cacheOnly'
 
 const cacheControl = {
     // Ignore headers, use uri as cache key, fetch only if not in cache.
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -23,15 +23,13 @@ const resizeMode = {
     center: 'center',
 } as const
 
-export type Priority = 'low' | 'normal' | 'high'
-
 const priority = {
     low: 'low',
     normal: 'normal',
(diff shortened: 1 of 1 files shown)


radko93:
@DylanVann this is minor but will help a lot!

krisztiaan-uic:
#676 is the same but with derived typing from the declared const

krisztiaan-uic:
@kulyk would you mind merging #676 into yours, so I could close that one?
also, anyone can just make these changes locally, and use `patch-package` to apply it in your current project until this one is merged

kulyk:
@krisztiaan-uic sure, thanks a lot! I'll do it on weekends and post and update here

kulyk:
Hi @krisztiaan-uic  I've merged #676 code into this branch, so now `Priority` and `Cache` types are derived from const values and exported. Thank you for your work!

However, all the recent CI checks fail due to Node v13 breaking change. I'll open a new PR to fix the CI checks a bit later

kulyk:
CircleCI fix is ready at #683 
@DylanVann hi, please take a look 🙏

DylanVann:
Hi @kulyk, thanks for the PR!

I merged the other one to fix the Cache type. I think it would be better to just use string unions rather than exporting enum like objects, then this issue won't end up reoccurring.

kulyk:
Hi @DylanVann,

The best thing is to have a new release :) Thank you for your time!

PR #676, opened 2020-04-29
fix(types): add derived types for priority and cache, export them

krisztiaan-uic:
Make types for `Cache` and `Priority` derived types from the corresponding `const`.
Export these as well, to simplify wrappers and further typing, thanks @vieiralucas

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -23,7 +23,7 @@ const resizeMode = {
     center: 'center',
 } as const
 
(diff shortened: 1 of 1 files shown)


vieiralucas:
@krisztiaan-uic Can you export these types too? So we can close #678 with your PR

krisztiaan-uic:
included in #654

PR #678, opened 2020-05-05
Export ResizeMode and Priority types

vieiralucas:
Thanks for this awesome library.

I think that having access to those types are important so users can make wrappers and accept them on their own components maintaining type safety.

I myself am currently needing this.


DylanVann:
LGTM!

PR #680, opened 2020-05-07
fix(android): make center ResizeMode work correctly

connoremma:
Howdy,

First of all great library - Second, I noticed that `ResizeMode.center` works differently on iOS and Android. iOS has the desired behavior whereas android is using a different `ScaleType`

I checked out the source code for one of Facebook's unit tests on Android for resize mode stuff and saw that they were using `ScaleType.CENTER_INSIDE` – this library is using `ScaleType.CENTER` 

Making this simple change fixed the behavior - LMK if any other info is needed in order to merge this

Thanks in advance


DylanVann:
LGTM, thank you!

PR #683, opened 2020-05-17
Use Node.js LTS version on CI (fixes Circle CI builds)

kulyk:
`yarn build` on Circle CI fails with ` Cannot find module '@babel/compat-data/corejs3-shipped-proposals'` error. This error appears in [unstable Node.js versions](https://github.com/nodejs/node/issues/32852).

Changed Circle CI config to use LTS Node.js version

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2
 jobs:
     build-and-test:
         docker:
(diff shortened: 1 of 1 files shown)


DylanVann:
Thank you! I switched the project to using GitHub actions, so we don't need to do this, and the error is fixed.

PR #688, opened 2020-06-03
fixes wrong cache type

lgenzelis:
Cache type for typescript was wrong. The typing for flow is the right one.


kulyk:
Hi, already made that at #654 
Also, all the CircleCI checks fail now, fixed that at #683

Issue #690, opened 2020-06-05
[iOS] Headers for one request are sent for all subsequent requests, even if no headers specified


andreialecu:
**Describe the bug**

When an image with `source={{uri: "...", headers: { host: "somehost.com", authorization: "..." }}` is loaded, all other subsequent images will use the same headers.

This can lead to images not showing, because the `host` header is wrong, and web servers will return 404 errors for the request.

**To Reproduce**
Steps to reproduce the behavior if possible, or a link to a reproduction repo:
1. Go to https://hookbin.com/ and create a new endpoint
2. Add an image with `source={{ uri: "hookbin endpoint #1", headers: { foo: "bar" } }}`.
3. Create another endpoint at hookbin
4. Add another image with `source={{ uri: "hookbin endpoint #2" }}`, ensure it loads after image 2.

The request for image on endpoint `#2` will include the `foo`: `bar` headers.

This is probably the reason of a lot of errors with images not properly showing in this repository.

**Additionally, this is a serious security issue, as signing credentials or other session tokens could be leaked to other servers. See screenshot below.**

**Expected behavior**

Headers should not be reused.

**Screenshots**

<img width="428" alt="image" src="https://user-images.githubusercontent.com/697707/83863179-aa406280-a72b-11ea-8fc6-c1c44218c3ac.png">

Notice how headers are leaked to other servers. The headers were only supposed to be used for images from the s3 bucket, but they are attached to every request.

**Dependency versions**

- React Native version: 0.62.1
- React version: 16.11.0
- React Native Fast Image version: 8.1.5

andreialecu:
Issue acknowledged by upstream SDWebImage contributor @dreampiggy at:

https://github.com/SDWebImage/SDWebImage/issues/3031#issuecomment-639376964

dreampiggy:
Not the issue of upstream dependency, this is a bug of react-native-fast-image implementation. Should use the correct API to do **Per-image-level-header-setup**.

andreialecu:
Fix is available at #691

andreialecu:
Closing since #691 was merged, and the fix is in version 8.3.0 as per https://github.com/DylanVann/react-native-fast-image/pull/691#issuecomment-659864093

PR #691, opened 2020-06-05
feat(ios): allow for for per-image-request-headers

### LINKED ISSUE #242: Some images are not showing.
First, Below is showing well.

```
<FastImage
              style={s.image}
              source={{
                uri: 'https://unsplash.it/400/400?image=1',
                headers:{ Authorization: 'someAuthToken' },
                priority: FastImage.priority.normal,
              }}
              resizeMode={FastImage.resizeMode.stretch}
            />
```

But, Below is not showing.

```
<FastImage
              style={s.image}
              source={{
                uri: 'https://s3.ap-northeast-2.amazonaws.com/zzim-app-images/intro/180529_1.png',
                headers:{ Authorization: 'someAuthToken' },
                priority: FastImage.priority.high,
              }}
              resizeMode={FastImage.resizeMode.stretch}
            />
```

Why can't I see any image? 
Can't I load image from s3?

So I test to upload first image in s3. 
1. Download : "https://unsplash.it/400/400?image=1"
2. Uri change : : "https://s3.ap-northeast-2.amazonaws.com/zzim-app-images/test.jpeg)"

But it's not showing too.

dreampiggy:
See detail issue description in #690. See SDWebImage issue: https://github.com/SDWebImage/SDWebImage/issues/3031
Close #242 

Fix the implementation on iOS for per-image-request-header-setup. Should not use `SDWebImageDownloader` which manage the global shared headers.

Should Use the context option of request modifier. Or when concurrent request comes, the headers may mass up.

This MR also bump webp coder because the old version contains OOM issue.

CC @DylanVann @andreialecu


andreialecu:
🎉 

(btw, I think you meant to link #690)

dreampiggy:
Extra changes:
1. Fix the issue of `data:image` for WebP images. Using SDWebImage's API can keep the external format support, including WebP/APNG/AVIF/HEIF. As well as iOS system default format.

dreampiggy:
The CI environment seems broken (Not related to this MR changes). I can ensure the build sucess on my local development environment.

image

image

dreampiggy:
@andreialecu If your RN iOS use CocoaPods. You can use the git commit dependnecy to test the behavior.

```
pod 'react-native-fast-image', :git => 'https://github.com/dreampiggy/react-native-fast-image.git', :branch => 'bugfix_ios_sd_header_per_image_request'
```

andreialecu:
I generally like to use [patch-package](https://www.npmjs.com/package/patch-package). I think it should work just the same.

I seem to be getting this error on `pod install` though. I can probably update the deployment target, but it may result in breaking changes for users?

```ts
[!] CocoaPods could not find compatible versions for pod "SDWebImage":
  In snapshot (Podfile.lock):
    SDWebImage (= 5.7.0, ~> 5.0)

  In Podfile:
    RNFastImage (from `../node_modules/react-native-fast-image`) was resolved to 8.1.5, which depends on
      SDWebImage (~> 5.8)

Specs satisfying the `SDWebImage (= 5.7.0, ~> 5.0), SDWebImage (~> 5.8)` dependency were found, but they required a higher minimum deployment target.
```

andreialecu:
Okay, deleting `Podfile.lock` then running `pod repo update` then `pod install` got it going.

It's probably an artifact of using patch-package to update the package.

andreialecu:
I seem to be seeing this crash:

image

Not sure what is wrong, could be a problem on my end.

andreialecu:
No luck:

```ts
  return (<FastImage
    style={{ height: 100, width: 100 }}
    source={{ uri: "https://dummyimage.com/600x400/000/fff"}}
  />);
```

Enters error handler here:

image

But more interestingly:

```ts
  return (<FastImage
    style={{ height: 100, width: 100 }}
    source={{ uri: "https://loremflickr.com/320/240"}}
  />);
```

This crashes with EXC_BAD_ACCESS shown in a previous message. I suspect it's because the URL is a redirect and doesn't store the image directly.

Both images display properly if I replace `FastImage` with the react-native `Image`

andreialecu:
Commenting out this line makes the images load:

https://github.com/DylanVann/react-native-fast-image/pull/691/files#diff-c18dcaa41187abeeba66a03f2fd7b89aR193

But that would make custom headers not work.

I guess something is incomplete related to the context.

dreampiggy:
Please paste the demo code or project so I can have a debug check tomorrow. Seems some more strange behavior inside.

andreialecu:
@dreampiggy minimal repro at: https://github.com/andreialecu/rnfastimage-691

To get it going, run:
```
yarn
cd ios && pod install
yarn ios
```

Take a look at `App.js`, I left a comment there. Should show both the crash and the empty images.

dreampiggy:
After using native debug code, I found the following issue:

1. Your example URL `https://dummyimage.com/600x400/000/fff`, use the experimental `multipart/mixed` contents, which is not considered in currernt SDWebImage implementation.
But React-Native provide the implemenation, see: https://github.com/facebook/react-native/blob/0.63-stable/React/Base/RCTMultipartDataTask.m#L101-L128

So, which means, until we support the same code, you can not request the stream-style image server provider.

2. Your example URL `https://dummyimage.com/600x400/000/fff` which use a not-trusted SSL certificate. (You have wrong Let's Encrypt config?). Because of this, you will trigger the SSL authoriation error for SDWebImage. Use `SDWebImageDownloaderAllowInvalidSSLCertificates` can skip this validation. But however, `react-native-fast-image` does not have anything customizable for this behavior. For temp workaround, I can hardcode this to always skip.

dreampiggy:
The `multipart/mixed` case I can create a Feature Request on SDWebImage side. This must need Client Code update to handle.

dreampiggy:
Send TODO feature reuqest: https://github.com/SDWebImage/SDWebImage/issues/3033

andreialecu:
Strange. I don't see any SSL certificate error in Chrome or iOS Safari on neither of the images. 

Could you clarify which one has the SSL issue?

andreialecu:
I've updated the code as per your latest commit and the second image still crashes the app. (loremflickr)

I don't think there's any SSL issue, because the crash occurs when using a google maps photos url as well. The only thing in common is that both do a redirect.

andreialecu:
Additionally, could you clarify where you see `multipart/mixed`? The raw headers for `https://dummyimage.com/600x400/000/fff` are:

```
$ curl https://dummyimage.com/600x400/000/fff -i

HTTP/2 200 
server: nginx
date: Sun, 07 Jun 2020 09:53:34 GMT
content-type: image/png
content-length: 1783
cache-control: public, max-age=7776000
expires: Sat, 05 Sep 2020 09:08:05 +0000
last-modified: Sun, 07 Jun 2020 09:08:05 GMT
x-srcache-fetch-status: HIT
x-srcache-store-status: BYPASS
access-control-allow-origin: *
access-control-allow-methods: GET, POST, OPTIONS
access-control-allow-headers: DNT,User-Agent,X-Requested-With,If-Modified-Since,Cache-Control,Content-Type,Range
access-control-expose-headers: Content-Length,Content-Range
```

Seems like all the data would be there. I see `content-type: image/png`

andreialecu:
```objective-c
    [self sd_setImageWithURL:_source.url
            placeholderImage:nil
                     options:options
//                     context:context
                    progress:^(NSInteger receivedSize, NSInteger expectedSize, NSURL * _Nullable targetURL) {
```

Again, I just verified that commenting out `context` here makes both images load perfectly. But the headers are (obviously) going to be missing.

migueldaipre:
I did the tests as described by you, and it worked perfectly with local Base64 WebP images.
I could only notice how the @andreialecu said, without commenting on the context the external images do not work.

I got the same error: EXC_BAD_ACCESS (code = 1, address = 0x0)
CFNetwork internal error

I wanted to understand more to help. If you can guide me.

I had problems with Flipper, I changed in PodFile to 0.37.0.

image

andreialecu:
Hey @dreampiggy, was just wondering if you saw the latest discussion here.

dreampiggy:
Busy working in daily job...

For the strange crash, I doubt there are something race condition for iOS native code to `mutable copy && copy the NSURLRequest`.

If I can reproduce the crash, maybe it's better to do not mutable copy and re-create a new NSURLRequest instead. I'll have a try and verify, if that can fix, I'll update the MR.

andreialecu:
Sure, I understand.

But there's not just the crash, neither of the images load with these changes unless context is removed. I think there's something else wrong with it.

dreampiggy:
Seems the fucky fact that `mutableCopy` is dangerous.

I write the exact same code in logic, to re-construct a new `NSURLRequest`, instead of mutable copy the original one and modify, the crash disappears. :)

image

dreampiggy:
@andreialecu 
Confirmed the bug because of SDWebImage's usage.

The `NSURLRequest.HTTPMethod` should not use nil, should use `GET` for default value.

I'll provide a hotfix on SDWebImage for this.

This PR already apply the patch, and can be mereged now.

ios/FastImage/FFFastImageView.m
```diff
@@ -142,12 +144,21 @@ - (void)reloadImage
...
+                [mutableRequest setValue:value forHTTPHeaderField:header];
+            }
+            return [mutableRequest copy];
         }];
+        SDWebImageContext *context = @{SDWebImageContextDownloadRequestModifier : requestModifier};
         
         // Set priority.
         SDWebImageOptions options = SDWebImageRetryFailed | SDWebImageHandleCookies;
+        // Bypass invalid SSL certificate, same behavior like React Native
+        options |= SDWebImageAllowInvalidSSLCertificates;
```
andreialecu, line 161: I wonder about this. This lib allows specifying HTTP headers, which would most likely be used for authentication. This option would allow a man-in-the-middle attacker to collect headers.

Should this be under a flag, as a separate PR/issue? 

I feel more thought is needed here. In my case, the changes in the PR work without this flag too.

andreialecu:
@dreampiggy awesome, I just tested it and it loads all images that were previously not loading, and headers are not leaking any more. There's one minor concern about skipping invalid SSL certs, which I added a comment about.

ios/FastImage/FFFastImageView.m
dreampiggy: I think react-native-fast-image should have a config via JS API for this. Current they have no such of thing.

The React Native already allows the un-trusted SSL error. I just do the same behavior to match them.

If you really don't want this, I can just revert.

dreampiggy: Try again. Maybe now some bad image server may not load the image. (However, React Native's `<Image>` can )

andreialecu: Do you have a link pointing out to this behavior in the RN Image component? I couldn't find anything.

dreampiggy: Seems I misunderstand the behavior. RN does not add the delegate to `URLSession:task:didReceiveChallenge:completionHandler:`. When self-host or some SSL error happaned, it will use iOS's default behavior.

However, this default behavior seems different from SDWebimage's code (which provide a default behavior when `SDWebImageAllowInvalidSSLCertificates` not passed).

Current keep this.

gwenoleR:
Hi guys, any news about this fix ?

dreampiggy:
I want to merge this, but seems not get any reply from the maintainer yet.

dreampiggy:
@DylanVann Welcome to see  you back here.
This PR not only just used for logic fix, but also fix the security issue talked about by #690. This means you leaks the HTTP headers like Cookie to other domain.

.gitignore
```diff
@@ -22,6 +22,10 @@ DerivedData
 *.xcuserstate
 project.xcworkspace
 
+# CocoaPods
+Podfile.lock
+Pods
```
dreampiggy, line 27: The `Podfile.lock` will cause massive conflict from different MR contributors.

Since we follows the sem version and this is used for example, ignore Podfile.lock is a better idea for avoid this. See SDWebImage's itself project maintain guideline.

DylanVann: If we don't have a Podfile.lock then we can't have reproducible builds / tests. Although this project does not have testing of the native side at the moment anyways...

I'd be alright not checking it in until native testing is implemented.

Thanks for updating this, looks great!

DylanVann:
@dreampiggy Thanks so much for implementing this. This was an aspect of SDWebImage that always frustrated me, great that it's fixed.

dreampiggy:
> @dreampiggy Thanks so much for implementing this. This was an aspect of SDWebImage that always frustrated me, great that it's fixed.

Note this MR also add the support for any external image format support on iOS. So maybe it's better to update the release changelog as well.
Like WebP/AVIF/SVG/PDF, etc.
The iOS user just need to register the coder plugin follow [SDWebImage's guide](https://github.com/SDWebImage/SDWebImage/wiki/Coder-Plugin-List), then all things done.

PR #698, opened 2020-06-25
fix(android): remove explicit use of UI thread

pawan52tiwari:
No need to send events via UI thread. This causes app to use main thread unnecessary that may result in an ARN exception


pawan52tiwari:
@DylanVann Can you please review this pull request

pawan52tiwari:
> @DylanVann Can you please review this pull request

@DylanVann  Can someone please review this cr

DylanVann:
LGTM!

PR #703, opened 2020-07-06
feat: animated WebP pausing

kaniaorzel:
In one of our projects, we need some tweaks connected with Animated WebP. 
This is changes in our fork, maybe somebody gonna need it, in the future :)

If you have questions or comments, feel free to write it in this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,8 +120,8 @@ Headers to load the image with. e.g. `{ Authorization: 'someAuthToken' }`.
 
 ### `source.isPaused?: string`
 
--   `false` **(Default)** - Animated Webp is started.
--   `true` - Animation Webp is stopped.
+-   `false` **(Default)** - Animated WebP is started.
+-   `true` - Animated WebP is stopped.
 
 ---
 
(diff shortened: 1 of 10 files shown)


dreampiggy:
For iOS level control. You can just use [SDWebImageDecodeFirstFrameOnly](https://sdwebimage.github.io/SDWebImage/Enums/SDWebImageOptions.html#/c:@E@SDWebImageOptions@SDWebImageDecodeFirstFrameOnly) to not load any Animation at all.

If you need dynamic control (like a callback, you can control when to start or stop), use the UIImageView(SDAnimatedImageView override these methods): isAniamting startAnimating stopAnimating. See [Apple's documentation](https://developer.apple.com/documentation/uikit/uiimageview/1621061-startanimating)

dreampiggy:
Seems this `isPause` is just not a binding type or something supports dynamic changes. For iOS, just simple, use 

```
if (source.isPaused) {
    options << SDWebImageDecodeFirstFrameOnly;
}
```

kaniaorzel:
Thank you @dreampiggy for feedback. I will look at it

PR #550, opened 2019-08-29
Support for defaultSource

danielgindi:
Both iOS and Android.
No support for URIs due to library limitations. It also does not make any sense. (As defaultSource should be available immediately, not for going to fetch from a remote location)

PR #708, opened 2020-07-17
BREAKING CHANGE: refactor enum types and remove static properties

DylanVann:
This PR modifies FastImage so that it no longer has static properties attached.
e.g. It does not have the properties priority, cacheControl, resizeMode, and preload.

The enum like properties (priority, cacheControl, and resizeMode) have been removed
in favor of using string union types, like React Native does.

preload has been added as a named export. So you can use "import { preload } from 'react-native-fast-image'.
Since this PR is already breaking I also removed the default export in favor of a named export.
You can now use "import { FastImage } from 'react-native-fast-image'".
There are lots of reasons to prefer named exports.


warrenalphonso:
@DylanVann When will you be merging this?

danielgindi:
> This PR modifies FastImage so that it no longer has static properties attached.
> e.g. It does not have the properties priority, cacheControl, resizeMode, and preload.
> 
> This PR also removes Flow support. It can be maintained somewhere else by interested parties.
> 
> The enum like properties (priority, cacheControl, and resizeMode) have been removed
> in favor of using string union types, like React Native does.
> 
> preload has been added as a named export. So you can use "import { preload } from 'react-native-fast-image'.
> Since this PR is already breaking I also removed the default export in favor of a named export.
> You can now use "import { FastImage } from 'react-native-fast-image'".
> There are lots of reasons to prefer named exports.

Would you mind first merging https://github.com/DylanVann/react-native-fast-image/pull/550.
It is used by many of your users. It's mature with more than a year behind it.
But your constant refactors makes it hard to re-re-base all the time.

PR #714, opened 2020-07-31
Adding loopCount and onAnimationComplete properties to FastImage for better gif/webp animation control.

sleeper-luke:
Pretty straightforward change overall, just hooking into the underlying native code in iOS/android to add the ability to specify a loop count for images (after which they will freeze on their final frame), and exposing a callback mechanism to inform react native when an animation has finished playing the specified number of loops.

A few notes:
1) Fast image relies on two other libraries for its native implementation (glide for android, SDWebImage for iOS). Glide does not have the ability to fire callbacks for each iteration of `loopCount`, and can only fire one when all loops have completed. So although SDWebImage _does_ support this, I ended up deciding to keep the behavior of the `onAnimationComplete` callback the same across platforms and just fire the event after `loopCount` is done.

2) The included `/dist` folder contains build files for the package that did not originally exist in the base `react-native-fast-image` repo. I imagine that there is some mechanism that builds these files through `npm` or `yarn` that I'm unaware of, because I couldn't figure out a way to omit these files from the repo and still have them build locally when the package was downloaded. (the `package.json` here does contains a `build` script that compile these files, but it has dependencies on other libraries that I didn't want to include with our app. Not sure how npm/yarn can build these without those dependencies, but they must be doing it somehow!) Can possibly revisit this later to see if we can remove this files from this repo. For CR purposes, feel free to ignore them.


cross19xx:
Any reason why this was closed?

PR #732, opened 2020-09-29
fix: xcode 12 compatibility

kasinskas:
Xcode 12 fails to build if a module depends on React instead of React-Core. Reference - https://github.com/facebook/react-native/issues/29633#issuecomment-694187116. Even though some packages might not be affected right now, it's still considered a [good change](https://github.com/facebook/react-native/issues/29633#issuecomment-694534795)


dylancom:
Need to get this merged.

PR #744, opened 2020-11-11
Update OnLoadEvent typings

SRandazzo:
This exposes the target property of the onLoadEvent.

Our use case for needing this is that when the image loads, we take the target (which is the react tag) and pass it over the bridge into Obj-C code so that we can grab a reference to the UIImageView and perform some manipulations.

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -54,6 +54,7 @@ export interface OnLoadEvent {
     nativeEvent: {
         width: number
         height: number
+        target: number
     }
 }
 


DylanVann:
I don't really understand the use case. Could you show some example code?

If this was done, would it make sense to do the same thing for other events?

DylanVann:
Thank you! Fixed in 8.6.11 (#1094). I made `target` optional, since Android with the legacy architecture doesn't send it.

PR #770, opened 2021-02-16
Add transition animation cross fade to the loading state

iagormoraes:
Hello, just added the transition property to allow Glide to make the animation cross fade to the image load state.


J6ey:
Why wasn't this merged?

PR #773, opened 2021-02-22
feat: move clearing logic to onDetachedFromWindow

WoLewicki:
Moved the logic of clearing views from `onDropViewInstance`, which is not the proper place for it since the view can be dropped, but still visible for during the transition (see https://github.com/software-mansion/react-native-screens/issues/797 and https://github.com/software-mansion/react-native-screens/issues/773). They have been moved to `onDetachedFromWindow`, which is the moment the view is not present in the window of the app. It required passing the `requestManager` to the `FastImageViewWithUrl` instance.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+## [8.6.3](https://github.com/DylanVann/react-native-fast-image/compare/v8.6.2...v8.6.3) (2022-10-31)
+
+
+### Bug Fixes
+
+* tintColor in fallback Image ([#882](https://github.com/DylanVann/react-native-fast-image/issues/882)) ([fbb6b68](https://github.com/DylanVann/react-native-fast-image/commit/fbb6b68a606f8f1fb9b570243d65bd410f21f63e))
+* use ColorValue type ([#939](https://github.com/DylanVann/react-native-fast-image/issues/939)) ([54376d8](https://github.com/DylanVann/react-native-fast-image/commit/54376d87f7e3826b884374a429c3879635dea246))
+
 ## [8.6.2](https://github.com/DylanVann/react-native-fast-image/compare/v8.6.1...v8.6.2) (2022-10-31)
 
 
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "react-native-fast-image",
-    "version": "8.6.2",
+    "version": "8.6.3",
     "description": "🚩 FastImage, performant React Native image component.",
     "keywords": [
         "cache",


ku8ar:
@WoLewicki  Are you sure this solution is OK? I have in app:

react-navigation v4
react-native-screens 2.16

and when I change screens plugged into bottomStackNavigator, then onDetachedFromWindow cleans images that are still on these screens.

Maybe we should plug in onDestroyView event here?

WoLewicki:
Can you post a reproduction here? I am not 100% sure if this solution is 100% correct.

ku8ar:
I have created a repro here in Example on master: https://github.com/ku8ar/react-native-screens

ff_broken

WoLewicki:
I changed the implementation to handle loading/clearing of the images in `onAttachedToWindow` and `onDetachedFromWindow`. Can you check if it fixes the issues? Or maybe I am missing something here too?

ku8ar:
When image is generated in `onAttachedToWindow`, it is not possible to dynamically change source. We need to recreate a new fast-image object. Right?

WoLewicki:
Yeah, we should probably keep the logic of loading in setting the source too. I will change the implementation to an easier one and I will ping you then.

scvgoe:
any progress on this?

WoLewicki:
I added loading the image on setting the source, can you test if all is ok now @ku8ar @scvgoe?

ku8ar:
@WoLewicki  checked. Looks ok. Thanks!

TheEhsanSarshar:
can you please merge it @DylanVann

chr4ss12:
bump, would be good to have this merged

moulie415:
@WoLewicki  Having flickering issues with this fork, here's some example code

`
import React from 'react';
import {NavigationContainer} from '@react-navigation/native';
import {createMaterialTopTabNavigator} from '@react-navigation/material-top-tabs';
import {ScrollView} from 'react-native';
import Image from 'react-native-fast-image';

const Tab1 = () => {
  return null;
};

const Tab2 = () => {
  return (
    <ScrollView style={{margin: 10}}>
      <Image
        style={{
          width: 600,
          height: 600,
        }}
        source={{
          uri: 'https://mht.wtf/post/content-aware-resize/sample-image.jpeg',
        }}
      />

      <Image
        style={{
          width: 600,
          height: 600,
        }}
        source={{
          uri: 'https://encrypted-tbn0.gstatic.com/images?q=tbn:ANd9GcT_12mZYw9c0QfCb875croxaX5vufbLRdrXLRhudSwVmHtuSKYOPise1ZTKQgl-_QzGdJc&usqp=CAU',
        }}
      />
    </ScrollView>
  );
};

const Tab3 = () => {
  return null;
};

const App = () => {
  const Tab = createMaterialTopTabNavigator();

  return (
    <NavigationContainer>
      <Tab.Navigator initialRouteName="Tab2">
        <Tab.Screen name="Tab1" component={Tab1} />
        <Tab.Screen name="Tab2" component={Tab2} />
        <Tab.Screen name="Tab3" component={Tab3} />
      </Tab.Navigator>
    </NavigationContainer>
  );
};

export default App;
`

if you test going back and forth from tab1 and tab2 you'll notice that the first image rendered always flickers

WoLewicki:
@moulie415 I tested the code and did not spot any flickering. Could you provide a video where it is clearly visible? Also, you are using `createMaterialTopTabNavigator` which does not use `react-native-screens` underneath, so there may be something connected to this package that does not work well with this PR.

moulie415:
> @moulie415 I tested the code and did not spot any flickering. Could you provide a video where it is clearly visible? Also, you are using `createMaterialTopTabNavigator` which does not use `react-native-screens` underneath, so there may be something connected to this package that does not work well with this PR.

@WoLewicki  I've managed to repro with both bottom and material top tabs

https://user-images.githubusercontent.com/10909374/129879290-aef6f282-9855-431d-b16d-8a317bc077a0.mp4

https://user-images.githubusercontent.com/10909374/129879274-06fe04fa-a05f-4600-b6dd-0d85c3e587ef.mp4

WoLewicki:
Hmm it is probably caused by the fact that the whole tab is detached from the native view hierarchy when the tab is changed with `react-native-screens` enabled. Can you spot flickering when you don't use the `react-native-screens` in this navigator (set https://reactnavigation.org/docs/bottom-tab-navigator/#detachinactivescreens to `false`)?

moulie415:
@WoLewicki yeh that works however  "react-native-fast-image": "^8.3.7" is fine regardless of what you use.

WoLewicki:
Yeah, I don't know an easy way to fix the bug and make it work correctly. I am open to suggestions.

chr4ss12:
@WoLewicki 

What about this? I know too little about how things work but this seemed to fix the original issue without breaking the other one

```
    @Override
    public void onDropViewInstance(final FastImageViewWithUrl view) {
        // This will cancel existing requests.
        if (requestManager != null) {
            runOnUiThread(
                new Runnable() {
                @Override
                public void run() {
                    requestManager.clear(view);
                }
            });
        }


      ....
```

essentially running the requestManager.clear() on UI thread

WoLewicki:
@chr4ss12 I also have not enough knowledge about this, but if it works, please make a PR with it and link it here so people can easily test it. Maybe it is a better solution to the problem.

PR #774, opened 2021-02-26
Android preload method null safety

BLANKAXU:
add null check to avoid crash


putuoka:
This is trap made error to me

````
 if (source != null && source.hasKey("uri") && !isNullOrEmpty(source.getString("uri"))) {
                                                                   ^
  symbol: method isNullOrEmpty(String)
Note: Some input files use or override a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
````

DylanVann:
Thank you! Fixed in 8.6.8 (#1091).

PR #781, opened 2021-03-25
Add instructions to install for iOS

AngelTrinidad:
With this instruction will finish correctly the installation process for iOS and prevent errors when try to use the component (like "FastImageView" was not found in the UIManager)


DylanVann:
Thank you!

PR #782, opened 2021-03-30
fix: make sure not to crash when source is invalid

### LINKED ISSUE #407: Crash on Android if source is an invalid URL
If I pass as `source` an object like this `{ uri: 'INVALID_URL' }` I get the following error on Android:
```
Error while updating property 'source' of a view managed by: FastImageView

null

Must not be null or empty
```
At the moment I am normalising the source using this (ugly) code to avoid the crash:
```ts
const normalisedSource = source && typeof source.uri === 'string' && !source.uri.split('https://')[1] ? null : source;
```
but I think it would be better if the library prevents it.

AmauryLiet:
Hello, found an issue on android, causing a crash if the uri is invalid.

Suggested solution: trigger onError instead

To reproduce, you can use: `source={{ uri: '/' }}`

Fixes #407


DylanVann (approved):
LGTM 🚢

PR #783, opened 2021-04-12
added mavenCentral() for gradle builds

maheshwarimrinal:
As jCenter is shutting down so adding mavenCentral() so that gradle builds fail


DylanVann:
Done on main. Thank you!

PR #785, opened 2021-04-18
Feature clear cache memory and disk

williardx:


evanjmg:
why was this deleted?

PR #787, opened 2021-05-05
feat(ios): add cancel image downloading when image remove from render

nomi9995:
add cancel image downloading when `isEnable` will `false` and start image loading when `isEnable` will `true`. you can see below example

# Example
```
const [isEnable, setIsEnable] = useState(true);

{isEnable && (
        <FastImage
          style={{width: 400, height: 400}}
          source={{
            uri: "https://upload.wikimedia.org/wikipedia/commons/f/ff/Pizigani_1367_Chart_10MB.jpg",
            priority: FastImage.priority.normal,
          }}
          resizeMode={FastImage.resizeMode.contain}
        />
      )}
```

#### Note: this feature is already implement in android but ios was missing this feature
ref: https://github.com/DylanVann/react-native-fast-image/blob/0439f7190f141e51a391c84890cdd8a7067c6ad3/android/src/main/java/com/dylanvann/fastimage/FastImageViewManager.java#L62

### DIFF
--- a/ios/FastImage/FFFastImageView.m
+++ b/ios/FastImage/FFFastImageView.m
@@ -1,5 +1,6 @@
 #import "FFFastImageView.h"
 #import <SDWebImage/UIImage+MultiFormat.h>
+#import <SDWebImage/UIView+WebCache.h>
 
 @interface FFFastImageView()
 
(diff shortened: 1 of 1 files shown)


DylanVann (approved):
I hadn't realized SDWebImage didn't implement this natively. Nice catch. 🚀

PR #811, opened 2021-08-16
Include `enableDiskCaching` static method

bewannabe96:
On older IOS devices with low memory often face app crash "out of memory" issue.
This gives an option to choose caching on disk instead of memory to avoid such problem

simply added `enableDiskCaching` static method to the package

### DIFF
--- a/android/src/main/java/com/dylanvann/fastimage/FastImageViewModule.java
+++ b/android/src/main/java/com/dylanvann/fastimage/FastImageViewModule.java
@@ -53,4 +53,9 @@ class FastImageViewModule extends ReactContextBaseJavaModule {
             }
         });
     }
+
+    @ReactMethod
+    public void enableDiskCaching() {
+        // does nothing on android
+    }
 }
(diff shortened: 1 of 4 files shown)


tarouboy:
Is it possible to enable disk caching for Android as well?

bewannabe96:
> Is it possible to enable disk caching for Android as well?

From my experience,android didn't lack of memory using react-native-fast-image. Long story short, this pr only make changes on iOS

DylanVann:
I generally try to merge things only when there's an implementation for both platforms. One of the really annoying things about RN development in the past has been features only working on one platform.

10000multiplier:
@bewannabe96 how should we clear the disk cache?

Zeroverss:
> I generally try to merge things only when there's an implementation for both platforms. One of the really annoying things about RN development in the past has been features only working on one platform.

But why we need common solution for both platform? Glide and SDWebImage are different libs
we can setup cache for android using cache control, throught setup RequestOptions
to do it for ios need to implement custom request image method, looks like more work instead of just setup SDWebImage config

PR #826, opened 2021-09-20
Update CI and (re)enable coverage

friederbluemle:
A couple of commits here:

1. Update and simplify CI script (e.g. replace deprecated bash script with `codecov-action`)
2. Run `yarn test --coverage` to generate and publish coverage data

Also another important improvement: Prevent running `yarn dv-scripts ci` on forks. Obviously, the step will fail in forked repos, as the secrets are not available in that context and of course you wouldn't want to publish from forks anyway.

Example of successful build/coverage report:

https://github.com/friederbluemle/react-native-fast-image/runs/3645227154
https://codecov.io/gh/friederbluemle/react-native-fast-image/tree/a0808cb138ed9e71fd40f1e21dee6dbec52042dd


.github/workflows/ci.yml
```diff
@@ -0,0 +1,18 @@
+name: ci
+on: [push]
```
DylanVann, line 2:
```suggestion
on: [push, pull_request]
```

It would be good to see that PRs are passing, I think this would make it work.

DylanVann:
Thanks, this looks great!

.github/workflows/ci.yml
friederbluemle: Oh.. actually now I remember why I didn't add it. PR builds also run in the context of the source repo, but don't contain the secrets (e.g. `NPM_TOKEN`), for obvious security reasons. Also we shouldn't release with every new PR build (`yarn dv-scripts ci`).

friederbluemle: I addressed this by changing the conditional to:

```
if: github.event.repository.fork == false && github.event_name != 'pull_request'
```

DylanVann: I think I have the rethink how `dv-scripts` works a bit to simplify this 😆 . Good for now though!

PR #843, opened 2021-10-30
[Android] Fix ClassCastException, TintContextWrapper cannot be cast to ThemedReactContext

thanakij:
I tried to tackle #840 due to `ClassCastException` in `setSrc()` method.

```
Context -> ContextWrapper -> ReactContext -> ThemedReactContext
Context -> ContextWrapper -> TintContextWrapper
```

So `TintContextWrapper` cannot be casted to `ThemedReactContext`. The two have `ContextWrapper` as a common base class.

This pull request introduces a logic to check if the object is an instance of `ThemedReactContext` before casting.


thanakij:
Hi @DylanVann, can you have a look into this?

thanakij:
According to [this discussion](https://stackoverflow.com/questions/37196284/android-support-library-23-4-0-android-support-v7-widget-tintcontextwrapper-can), ContextWrapper's `getBaseContext()` can be used for extracting `Activity` and `ReactContext`. 

As a result, I have updated my PR so this time, not only it will prevent CastCastException from the current codebase but also try its best to recover `ReactContext`!

PR #872, opened 2022-02-04
fix: android etag caching

hannojg:


hannojg:
Sorry, that was meant to go against my fork!

PR #876, opened 2022-02-18
Added fade transition for Glide and SDWebImage

J6ey:


mehthabux:
working great!!! i used this PR

hirbod:
While I think this is a great feature, it should be controlled with a prop `withCrossfade` or `fade`. What do you think @DylanVann? Do you need some support here?

PR #906, opened 2022-06-24
Support animated webp

452MJ:


KishorJena:
is it working with latest RN?

PR #927, opened 2022-09-09
Add option to disable interpolation when resizing small images

### LINKED ISSUE #926: Add option to disable interpolation when resizing small images
When working with pixel art, interpolation needs to be disabled otherwise the image looks blurry. Example:

With interpolation:
image

Without:
image

Image in reference is https://raw.githubusercontent.com/PokeAPI/sprites/master/sprites/pokemon/1.png

There should be an option to disable this (especially because React's Image cannot do this)

FancyFurret:
Fixes #926

I opted to add pixelPerfect to source, since it makes the android implementation much easier. It may make more sense to add it to `FastImageProps`, but then I'm not sure how we would access it from Android's `getOptions`. (Since downsampling/compression/transforming need to be disabled)

It works on my iOS device and sim, and my Android emulator (do not have a physical Android device). I'm still pretty new to React Native, Android, and iOS development so let me know if you see anything wrong!

### DIFF
--- a/android/src/main/java/com/dylanvann/fastimage/FastImageViewConverter.java
+++ b/android/src/main/java/com/dylanvann/fastimage/FastImageViewConverter.java
@@ -13,6 +13,7 @@ import com.bumptech.glide.Priority;
 import com.bumptech.glide.load.engine.DiskCacheStrategy;
 import com.bumptech.glide.load.model.Headers;
 import com.bumptech.glide.load.model.LazyHeaders;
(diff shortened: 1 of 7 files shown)


DylanVann:
Nice job! I'm not sure `pixelPerfect` is the best name for this prop. Earlier in the development of this library I thought it would be best to expose the same API for iOS and Android. As time goes on though I'm thinking it's just not always possible. It may be better to expose:

- `androidDownsampleStrategy`
- `iosMinificationFilter`
- `iosMagnificationFilter`

PR #933, opened 2022-09-27
[WIP] Feat: migration to RN new architecture

bhimanshu-kalra:


hirbod:
This PR would install everything around Fabric even when you don't enable it. All flags to prevent this are missing.

Issue #865, opened 2022-01-19
[Android] `onLoad` will never get fired when height and width set to 0 on


matinzd:
**Describe the bug**
`onLoad` will never get fired when height and width set to 0 on android

**To Reproduce**
Steps to reproduce the behavior if possible, or a link to a reproduction repo:
1. Set the initial value of height and width of FastImage component to zero
2. Then wait for loading
3. onLoad never get fired

**Expected behavior**
`onLoad` should get fired even with initial zero width and height. 

**Screenshots**

**Dependency versions**

- React Native version: 0.66
- React version: 17
- React Native Fast Image version: 8.5.11

demedos:
Related: 
- https://github.com/DylanVann/react-native-fast-image/issues/687
- https://github.com/DylanVann/react-native-fast-image/issues/608
- https://github.com/DylanVann/react-native-fast-image/issues/562
- https://github.com/DylanVann/react-native-fast-image/issues/529
- https://github.com/DylanVann/react-native-fast-image/issues/461
- https://github.com/DylanVann/react-native-fast-image/issues/391


https://github.com/DylanVann/react-native-fast-image/pull/446 should fix the issue but is open for almost 2 years now.
You can use [patch-package](https://github.com/ds300/patch-package) to fix the issue until it gets merged; you just have to manually apply the changes of the PR to the `node_modules/react-native-fast-image` files and run 
`npx patch-package react-native-fast-image`.
It is advised to delete the `node_modules/react-native-fast-image/build` folder before running patch-package to avoid generating a huge diff.

sajin321:
thanks @demedos  for specifying the fix, i have gone through the fix #446 but seems source is changed after that. applied fix of target size on view converter class , it is working fine now. tested on android 11 and 10. don't know much about its impact in other dependencies. i am attaching the code of patch file ,guys please review it.


```
diff --git a/node_modules/react-native-fast-image/android/src/main/java/com/dylanvann/fastimage/FastImageViewConverter.java b/node_modules/react-native-fast-image/android/src/main/java/com/dylanvann/fastimage/FastImageViewConverter.java
index d86f66f..cfd7f60 100644
--- a/node_modules/react-native-fast-image/android/src/main/java/com/dylanvann/fastimage/FastImageViewConverter.java
+++ b/node_modules/react-native-fast-image/android/src/main/java/com/dylanvann/fastimage/FastImageViewConverter.java
[...]

PR #446, opened 2019-04-12
android: fix onload not called

cctanfujun:

PR #953, opened 2022-12-15
fix: Android `onLoad` event when view width and height are zero

thomas-coldwell:
Hey @DylanVann 👋

This PR aims to fix an existing issue with the `onLoad` event not firing on Android when the view style's width or height is zero as documented here https://github.com/DylanVann/react-native-fast-image/issues/865. There is a similar PR https://github.com/DylanVann/react-native-fast-image/pull/446 that was opened, but as mentioned in reviews this causes crashes due to the `.override(Target.SIZE_ORIGINAL)` method being applied to Glide. Instead, this PR follows the recommended way of getting the natural image dimensions as documented here in the Glide repo https://github.com/bumptech/glide/issues/781#issuecomment-160953996.

The changes here include:
- Adds a second target of a Size class with the relevant decoder + transcoder to achieve this
- Aligns the `onLoad` method on Android to work in the same way iOS does - reporting the original natural image dimensions
- Some minor dev setup fixes to get the example app up and running

These changes are best tested with the `AutoSize` component in the example app - as you can see setting the width to zero still triggers the `onLoad` event and this example is then rendered correctly.

### DIFF
--- a/ReactNativeFastImageExample/android/build.gradle
+++ b/ReactNativeFastImageExample/android/build.gradle
@@ -1,5 +1,7 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 
+def REACT_NATIVE_VERSION = new File(['node', '--print',"JSON.parse(require('fs').readFileSync(require.resolve('react-native/package.json'), 'utf-8')).version"].execute(null, rootDir).text.trim())
+
 buildscript {
     ext {
         buildToolsVersion = "30.0.2"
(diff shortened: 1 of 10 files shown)


Beamanator:
@DylanVann Is there any chance you can let us know your thoughts on this change? @thomas-coldwell and I would be happy to discuss with you if you have any questions / concerns about this update

Beamanator:
Bump? 🙃

PR #959, opened 2023-01-09
docs: priority description

brennansaul:
Added the following description for the fast image `source.priority` prop. 

> Indicates the load order priority of an image. Images with `FastImage.priority.high` will load before images in a similar context with low or normal priority.


DylanVann:
Thank you! Included in #1096.

PR #973, opened 2023-02-21
null object makes app crash

kernelsrinivas:


android/src/main/java/com/dylanvann/fastimage/FastImageViewConverter.java
```diff
@@ -108,7 +108,7 @@ static RequestOptions getOptions(Context context, FastImageSource imageSource, R
                 .priority(priority)
                 .placeholder(TRANSPARENT_DRAWABLE);
 
-        if (imageSource.isResource()) {
+        if (imageSource != null && imageSource.isResource()) {
```
haresh12, line 111: would be good if we use optional chaining here like imageSource?.isResource()

DylanVann:
Thank you! Fixed in 8.6.8 (#1091).

PR #989, opened 2023-06-09
Add export to Cache type

ruslan-hontar:
In accordance with all `react-native-fast-image` types just add export to the `Cache` type


DylanVann:
Thank you! Fixed in 8.6.10 (#1093).

PR #992, opened 2023-06-30
set onLayout prop on wrapping view

praisedavid787:


DylanVann:
Thank you! Fixed in 8.6.12 (#1095).

PR #1000, opened 2023-07-28
Feat downsample

inckie:
Added new source properties to downscale large images `maxWidth`, `maxHeight`. If used, both are required.


inckie:
Did not meant to target parent repo, sorry.

PR #1018, opened 2023-11-24
chore: changed glide license url

cetfu:
when I wanted to look up the license for the glide library, I noticed that github couldn't find the license file. then I found that the correct file was "LICENSE" instead of "license". so I wanted to change that


DylanVann:
Thank you! Included in #1096.

cetfu:
OMG. Has react-native-fast-image made a comeback? @DylanVann 😯

PR #1019, opened 2023-11-29
fix: ios 17 support fixed

karamanmertcan:


karamanmertcan:
In IOS 17 not working without these changes.

crafterm:
Can confirm with our experience as well, we needed to apply this patch, otherwise had various missing views/strange behaviour in our app.

DylanVann:
Thanks @karamanmertcan! This landed in #1087 (8.6.6), with you as a co-author.

PR #1082, opened 2026-09-22
fix: prevent crash on zero or invalid image dimensions

SamPeterberg:
This PR fixes a critical iOS crash occurring in FFFastImageView.m when attempting to apply a tint color to an image with zero or invalid dimensions. The crash was caused by UIGraphicsBeginImageContextWithOptions receiving invalid size parameters and returning nil, leading to an unhandled exception.


DylanVann:
Thanks @SamPeterberg!

